PR #1, opened 2015-10-13
Build and install easy

valorkin:
added prepublish script and ignored generated files
`npm link` or `npm run build` will build all files needed before publishing


ocombe:
Oh, does it make on file of definitions that will work in a project when you include it ?
I'll try this tonight but that would be awesome, I didn't liked to publish the generated files either.

As for exporting the interface Observable, I'm not sure that's a good idea since there will be an official one in angular 2: https://github.com/angular/angular/blob/b96784756ced8ae5caf8ad9a3a61206372c23baf/modules/angular2/src/core/facade/async.ts#L94 (if @vsavkin doesn't forget to do it :D)

valorkin:
@ocombe definitions is fine, I am testing locally `npm link` + `npm install ../ng2-translate`
As for now you can not avoid exporting Observable, or you will have:
```bash
src/translate.service.ts(64,28): error TS4055: Return type of public method from exported class has or is using private name 'Observable'.
src/translate.service.ts(85,39): error TS4055: Return type of public method from exported class has or is using private name 'Observable'.
src/translate.service.ts(103,23): error TS4055: Return type of public method from exported class has or is using private name 'Observable'.
```

ocombe:
Ok, I guess it doesn't matter since you only import it if you want & it shouldn't override the official implementation. We'll update the lib when they make it available anyway.

ocombe:
So I tried this PR and it doesn't work for me. After installing it, it runs tsc, but in my file typings/tsc.d.ts it links to `/// <reference path="../node_modules/ng2-translate/index.d.ts" />` which doesn't exist since you deleted it.
Then when I try to build my local ts files, I get the following errors:
```
src\app\components\app.ts(6,32): error TS2307: Cannot find module 'ng2-translate'.
src\app\components\home.ts(5,29): error TS2307: Cannot find module 'ng2-translate'.
src\app\components\params.ts(4,29): error TS2307: Cannot find module 'ng2-translate'.
src\app\components\show.ts(12,29): error TS2307: Cannot find module 'ng2-translate'.
typings\tsd.d.ts(11,1): error TS6053: File 'C:/wamp/www/subNode/node_modules/ng2-translate/index.d.ts' not found.
```
And the js files are never built, which means that my app tries to load them but cannot and doesn't work.

valorkin:
Did not you forgot to `npm link` before `npm i path/to/ng2-translate` from other project?
but thing I missed indeed that you are using `.d.ts` hack in index.d.ts
```js
declare module 'ng2-translate' {
    import index = require("index");
    export = index;
}
```

ocombe:
Oh I didn't know `npm link`, I just installed it using the link to your fork (I think it does the same in the end)

valorkin:
sec, let me try one thing

valorkin:
can you try now?

ocombe:
So I had tones of error messages about duplicates, I had to remove everything from the tsd.ts file into the ng2-translate file. Then it compiled ok, but I couldn't load my app since it was missing the js files :)

valorkin:
tsc version? I am currently at home on mac, and it works differently then at work in linux
I have `1.6.2`

ocombe:
I have 1.6.2 too, but on Windows.
Anyway, Pawel should send me his notes for his talk at Angular Connect next week, I'll see if thee is a good solution for all of this and let you know.
Worst case, I'll have to wait for his talk next week :)

valorkin:
windows >.< WHY? :)
I have started to writing bash script:)
I can make docker file for test :)

ocombe:
Ahah because this is my computer at home, and I like to play on it as well. Also I can test on IE easily!
I can link you to my app if you wanna try

valorkin:
yes please, link me :)

ocombe:
https://github.com/ocombe/subNode/tree/2.0
branch 2.0
You should be able to start it with `gulp dev` once `npm install` is completed

valorkin:
node version btw?

ocombe:
Hmm I'm on 4.1.1 & npm 3.3.4

valorkin:
can not be! gulp-cli doesn't work with node 4 oO
https://github.com/gulpjs/gulp-cli/issues/9

ocombe:
It does with release 0.3.0 :)

ocombe:
Or not, I installed branch 4.0 to be exact https://github.com/gulpjs/gulp-cli/tree/4.0

valorkin:
ok, I got it
you actually compiling node_modules/**/*.ts files,
recompiling angular2 and all other dependencies may be not the best thing to do
screen shot 2015-10-14 at 01 36 27

valorkin:
don't pay attention to duplicates too much
```
npm i ../ng2-translate
tsd link
gulp dev
```
open in browser :)

ocombe:
Well it still tries to load the js files that don't exist

valorkin:
windows, what can I say
linux\osx works good >.<

ocombe:
I don't think that's the issue here :D
With the blog post provided by Brad, we might want to look at the angular repo to see how they include the typescript definitions

valorkin:
you can try to do it docker, I am not lying :)
and I will check on windows later
but yes, we should ng2 dist

robwormald:
new resolution stuff works like this:
- forget about `///ref` style stuff
- forget about `tsd` 
- compile your `.ts` source to commonsJS with the `--declaration` compiler option
- you want your `.d.ts` files to live in parallel to your `.js` files. 

either:
- reference `typings: "your/lib.d.ts"` in package.json
- include an `index.d.ts` file in the published npm distro.

angular2 is currently shipping the necessary ambient modules (es6-shim and whatnot) and support for that should improve in the next couple of TS releases. 

users can then consume your library by:
```typescript
//this will check in package.json, then for index.d.ts
import {AwesomeStuff} from 'ng2-translate' 
```
or if i want to pull out something deeper
```typescript
//this will look for a d.ts file at ng2-translate/deep/thoughts.d.ts
import {DeepThoughts} from 'ng2-translate/deep/thoughts`
```
and that should be all that's required :tada:

ocombe:
Nice, so I just need to add the typings parameter into package.json and that should be enough (the rest is already done). I'll try that :)

robwormald:
see #2 (sorry @valorkin, but TSD is out!) 

re: Observable type - that's my goal this week, to get that fixed before AngularConnect, so i would avoid messing with it too much - it'll be an export from angular2 you can just use.

valorkin:
@robwormald why sorry? it is awesome if it will work! :)

valorkin:
closed in favour of #2

PR #6, opened 2015-11-24
Add: Show key instead "" when no value in translate files

nglazov:
Fix: Remove tslint errors on build (semicolons, qoutes type, declaration order)


ocombe:
Thanks :)

PR #30, opened 2016-02-04
feat(cli): compatible with angular cli using 'ng install'

jkuri:
Hi @ocombe. I added bundler script to make it compatible with angular-clis `ng install ng2-translate` command. If you can merge this and publish to npm I think we can use this library as a real-world example as it also provides one provider in the [documentation](https://github.com/angular/angular-cli/pull/173) instead of a test library.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,12 @@ coverage
 !karma-test-shim.js
 *.map
 *.d.ts
+!make.js
+
+#################
+## Bundles
+#################
+bundles/
 
 #################
 ## JetBrains
--- a/make.js
+++ b/make.js
@@ -1,0 +1,36 @@
+var pkg     = require('./package.json');
+var path    = require('path');
+var Builder = require('systemjs-builder');
+var name    = pkg.name;
+
+var builder = new Builder();
+var config = {
+  baseURL: '.',
+  transpiler: 'typescript',
+  typescriptOptions: {
+    module: 'cjs'
+  },
+  map: {
+    typescript: './node_modules/typescript/lib/typescript.js',
+    angular2: path.resolve('node_modules/angular2'),
+    rxjs: path.resolve('node_modules/rxjs')
+  },
+  paths: {
+    '*': '*.js'
+  },
+  meta: {
+    'node_modules/angular2/*': { build: false },
+    'node_modules/rxjs/*': { build: false }
+  },
+};
+
+builder.config(config);
+
+builder
+.bundle(name, path.resolve(__dirname, 'bundles/', name + '.js'))
+.then(function() {
+  console.log('Build complete.');
+})
+.catch(function(err) {
+  console.log('Error', err);
+});
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "test": "tsc && karma start",
     "test-watch": "tsc && karma start --no-single-run --auto-watch",
     "commit": "git-cz",
-    "prepublish": "tsc",
+    "prepublish": "tsc && node make.js",
     "semantic-release": "semantic-release pre && npm publish && semantic-release post"
   },
   "repository": {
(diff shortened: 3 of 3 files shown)


ocombe:
Hey thanks, I meant to work on that this week end :-)

Just a quick question, why do you call tsc 2 times in the npm prepublish script?
`tsc && tsc -d && node make.js`

jkuri:
tsc -d creates correlated .d.ts files.

ocombe:
`tsc` does that as well since I defined `declaration` to true in tsconfig.json: https://github.com/ocombe/ng2-translate/blob/master/tsconfig.json#L9

jkuri:
Sorry, I didn't noticed. I removed it.

ocombe:
Perfect thanks, I'll test it tonight when I get back home :)

jkuri:
Okay, thanks. Please ping me when (if) you will publish this to npm.

ocombe:
This has been merged, and is available in 1.6.1.
But when I try to install the lib with the Angular CLI, I get the error:
```sh
the `ember generate <entity-name>` command requires an entity name to be specified
```

I'm not sure what wrong here :(

jkuri:
Thanks. Will test it right away. Did you use `angular-cli` master and linked correctly? npm package is a bit outdated at the moment.

jkuri:
Please remove `bundles/` from `.gitignore`, push and republish, that's what I did in one of my components today so the `bundles/` and correlated script is found by `ember-cli` npm script. 
Not sure why but when you install with `npm install ng2-translate` the `bundles/` is created correctly, but when use `ember-cli` npm script for installing the package the `bundles/` not comes with it and then package is not recognized as authentic. 
If you can do that I can test it right away. 
Btw, I re-checked this repository now and I think this is completely compatible with Angular CLI `ng install` command, but cannot be sure until I see it working.

jkuri:
Note for using `angular-cli` master:
````shell
git clone https://github.com/angular/angular-cli.git
cd angular-cli
npm install
npm link
````
````shell
ng new test-project
cd test-project
npm link angular-cli
ng install ng2-translate
````
Hope this helps.

**Comment by ocombe** [comment]
I just tried and now I have this:
```sh
ng install ng2-translate
Installing 3rd party package: ng2-translate...
Removing packages which were not recognized as authentic: ng2-translate 
No package to process. Quitting.
Done.
```

ocombe:
Maybe the problem is that I export my modules as commonjs and the bundle is systemjs ?

jkuri:
Please try removing bundles/ from .gitignore, then commit and push and then
republish to npm as 1.6.2. I have exact case today and it worked for me.

ocombe:
I did that already

jkuri:
Then you forgot to `git add`, not sure how `ember-cli`s npm script is connecting to github or npm registry, but you need to have `bundles/` directory and correlated `ng2-translate.js` in your repository. That worked for me. Also users will have direct access to your bundle.

ocombe:
Ok I found it, it fails because of that: https://github.com/angular/angular-cli/blob/b8dfe0c414c4b7eb20397236185f95cac4e7121a/addon/ng2/tasks/lib-install.js#L563
It checks that I have a .ts at the root, but I don't push it on npm (because it causes a problem with some IDEs that consider .ts instead of .d.ts files for typings).

jkuri:
I was just debugging this and found out the same...

jkuri:
This is the entry point and should be included in order to work, `lib-install.js` parses what library providers from this file.

ocombe:
I publish the .js file, but not the .ts file.
lib-install should be able to parse the .d.ts file instead if necessary.

jkuri:
Yes, you are right. I will make a fix asap. I never encountered on a problem that main .ts file not being included. Until then, can you just try if it's working with that file is included?

ocombe:
Ok I just released a new version that doesn't ignore .ts files for npm anymore, and it works now !
But, it doesn't import with the correct path:
`import {TranslateService} from 'ng2-translate';` should be `import {TranslateService} from 'ng2-translate/ng2-translate';`

Also it doesn't inject the provider in the bootstrap file even if I say yes...

jkuri:
@ocombe thanks. I just tested it and I can confirm this package is now 100% compatible. 

````typescript
import {TranslateService} from 'ng2-translate';
````
this is right, please look at https://github.com/ocombe/ng2-translate/blob/master/bundles/ng2-translate.js#L275

Also, provider gets injected properly in my case:
````typescript
import {TranslateService} from 'ng2-translate';

bootstrap(ProjectApp, [
  ROUTER_PROVIDERS, 
  TranslateService
]);
````

Thanks again. I will notify @NathanWalker who is responsible for documentation about 3rd party libraries to add this library as a real world example on angular-cli docs, if you don't mind?

ocombe:
Ok, no problem to add it to the docs, sure :)

PR #31, opened 2016-02-10
added simple directive

sclausen:
The translation key can be passed as an attribute, or the content of the element. It's also possible to pass an object with interpolation key-value pairs.


ocombe:
Hello, thanks for this PR, it's very complete and I like that !

I'm a bit worried about the fact that it can use the content of the element because of the risks of injection.
Also I think that it's not a good idea to use "nativeElement" because you cannot use webworkers: https://angular.io/docs/ts/latest/api/core/ElementRef-class.html

sclausen:
Yeah, you're totally right about nativeElement, but I didn't found out how to adequately manipulate the content from a `Directive` or `Component`. 
Could you explain your concerns regarding injection? Maybe, we can drop this feature because of this security problem alone not digging into the `nativeElement` issue, although I really like to have my translation keys as innerHTML.

ocombe:
Well you're using innerHTML, which means that you don't check the content of the element. In angular 1 we had [$sce](https://docs.angularjs.org/api/ng/service/$sce) that did that for us.
Let's say that your element contains the name of your users, one user changes his name to some js or sql code, and then you get the content with innerHTML and you try to use that as a key in the TranslateService & TranslateParser, I have no idea what could happen but it could be a problem somehow.

hpop:
There is not much documentation about the right way to manipulate the DOM, so I just took a look how the angular native directives do it. (for example [ngClass](https://github.com/angular/angular/blob/master/modules/angular2/src/common/directives/ng_class.ts)  )
This seems to be the right way:

```typescript
constructor(
  public translateService: TranslateService,
  public ref: ElementRef,
  public renderer:Renderer
) {
}

updateValue(key: string) {
  this.translateService.get(key, this.interpolateParams).subscribe((res: string) => {
    this.renderer.setText(this.ref.nativeElement, res);
  });
}
```

sclausen:
@hpop: but your code prevents using html in translations.

hpop:
Is html output necessary? As far as I understand it, there is no html support in pipes, so ng2-translate does't support html output right now. Or am I missing something?
Bonus: If we are not allowing html in translations, we don't have to build a implementation of $sce. :)

sclausen:
Apart from questioning html support, everything you wrote is correct. The thing is, I'm pretty certain, html support is essential.
You would have to translate every part of longer texts with markup or links separately which is in some languages not possible at all, because of different sentence structures.

hpop:
Ok. Good point.

NathanWalker:
If what we need is an Angular2 sanitizer to make this directive a reality, then I could make an Angular2 html sanitizer library based off this excellent one:
https://github.com/google/caja/blob/master/src/com/google/caja/plugin/html-sanitizer.js
?

Try it out here:
http://codepen.io/devilish/pen/aCJuH

There's also this alternative:
https://github.com/punkave/sanitize-html

If everyone is in agreement that having one would solve this and then integrating such a library here would allow this PR to be merged, I could create this lib this week.

sclausen:
@NathanWalker That would be great!I had a look into html-sanitizer and unfortunately it has some other dependencies to the caja tool :-/ I hope this is not a showstopper. On the other hand, sanitize-html is browserified and minified still 125K, so maybe resolving those dependencies in cajas html-sanizier would be a good start.

ocombe:
Maybe we could just add a note in the docs saying: "hey be careful" and go on with it ?

NathanWalker:
Be careful warning may suffice for time being. I think if a sanitizer is created, we could add a wiki here that described how to use the sanitizer with the library if desired. Maybe just take consideration in how this directive is implemented to allow easy plug/play with a 3rd party sanitizer :+1:

ocombe:
Yeah I think that's a better way to do that. I'll try to merge this tonight

src/translate.directive.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+  }
+
+  /**
+   * updates the translation
+   * @param  key
+   */
+  updateValue(key: string) {
+    this.translateService.get(key, this.interpolateParams).subscribe((res: string) => {
+      this.ref.nativeElement.innerHTML = res;
```
NathanWalker: I would be weary of this as this would not work in a `native` environment like NativeScript for instance. May be a way to let the `Renderer` handle this?

ocombe: yeah the PR is old and a lot of things have changed and need to be updated :)

NathanWalker: Maybe let this to go, and then I could make a {N} plugin that would provide a `native` alternative to this directive that could be provided for at runtime.
I wouldn't let it hold up a merge.

sclausen: If someone could help me with the `Renderer` approach, I would be really glad :D

ocombe:
Now that I've fixed the OnPush bug with the Pipe, do we still need this directive ?

NathanWalker:
@ocombe Awesome! Not sure we do need the directive anymore. @sclausen I like the additional flexibility it adds but may add unneeded complexity at this point since `OnPush` is now possible?

sclausen:
I would really enjoy to see my templates more like `<td translate>cool_translate_key</td>` than `<td>{{'cool_translate_key' | translate}}</td>`, but I think its just a matter of aesthetics. There is no real 'need' for this and to gain simplicity in code, we should omit this directive.

sclausen:
Oh, I totally forgot about the html thing. How can we achieve the rendering of html-translations (e.g. for formatted message bodies) without having the directive?

ocombe:
What are the html translations you're talking about ?

sclausen:
I thought of html translations e.g. for messages with variables.

The following could be the translation value:
``` html
<p>Dear {{username}},</p>
<p>Your order "{{order.name}}" has been shipped and will arrive est. {{order.estArrive}}</p>
<p>Sincerly<br />
Your [cool-company-name] Shipping Team
</p>
```
This could be the view:
``` html
<message-body translate-params="{order:order, username: username}" translate>
</message-body>
```

If this wouldn't be possible, it would be totally cumbersome to create multiple translation values to compose such a message and furthermore, not every language has the same sentence construction, so the variables are sometimes at a different place in the sentence.

Do you have a different solution for this problem?

sclausen:
@NathanWalker still interested in writing a sanitization lib for angular2?

NathanWalker:
@sclausen Yes! Once the issue with OnPush was solved for me, creating the sanitization library fell to bottom of my list though.

Here we go though, just created a shell for the lib here:
https://github.com/NathanWalker/ng2-sanitize

Invited you to collab :+1: 

I'd like to move forward with either of these options for the core sanitizer, lemme know what you think:
A. https://github.com/punkave/sanitize-html
or...
B. Port angular 1.x sanitizer to TypeScript and integrate: https://github.com/angular/angular.js/blob/master/src/ngSanitize/sanitize.js

ocombe:
The official sanitizer is real, we can use this: angular/angular/pull/8453

sclausen:
@ocombe angular/angular#8453 has been merged, so we just have to wait for rc2 to be released to :smile: what do you think should be done now? should I continue to work on this pull request and add the new sanitizer?

ocombe:
yes, I don't see why not, this was my only concern

ocombe:
You might want to pull the latest changes that I just pushed to make it work with RC2 :)

sclausen:
I also messed up the tests :-1: Im currently working on it ;)

sclausen:
@ocombe I desperately tried to get a test to work, but testing a directive with the depending translate service is too complicated for me. I tried my directive manually and it works, but I can't put a working test case for my directive together. Do you have any idea, how I can test this?

mchambaud:
It would be great to have this PR finalized.

Is there anything I can do to help?

sclausen:
@mchambaud beside the tests, there is still an issue with correct updates when switching the language. You could dig into both issues. Since I started this, angular got further and testing is now very different, so you could basically rewrite the whole test setup.

ocombe:
I'd like to see if we can use the i18n directive for this, maybe make a directive that could be provided instead of the existing i18n directive so that users can switch from the current i18n angular implementation to ng2-translate easily, and maybe even use the new ng-xi18n messages extractor

sclausen:
I still have the problem, that 
`<p translate>hello_world</p>` doesn't update after first language change, but `<p translate [translate-values]="{value:'my-name'}">greetings</p>` works properly.
One strange thing I discovered is, that `translate.get` doesn't return after the first language change in the first example. Has anybody an idea, why this happens?

deepu105:
@sclausen I know this may be very late but still why not `<p translate="my.trans.key"></p>` as its the format used in angular-i18n which is very famous for ng1 it would make migrations very easy. Also you might want to consider using a prefix to translate as its a HTML5 attribute and usually ends up having issues when using protractor

deepu105:
@sclausen sorry just saw the code, ignore my first comment. Seems you already support key as attribute value. May be you could consider adding a prefix to the `translate` directive. something like `ng-translate` may be

deepu105:
btw we currently use the below as a workaroud as I didnt want to write a complete directive from scratch. Works perfectly

```
import { Component } from '@angular/core';

@Component({
    selector: '[jhi-translate],[jhiTranslate]',
    template: '<span [innerHTML]="key | translate:args"></span>',
    inputs: ['key:jhi-translate', 'args:translate-values']
})
export class <%=jhiPrefixCapitalized%>Translate {

    private key: string;
    private args: any;
    //FIXME add support to pass translate-compile/ directives in translated content doesnt work
}
```

sclausen:
@deepu105 I don't think there is a need of prefixing the translate attribute-component with ng. When I created this, I had the compatibility to pascal precht's angular-translate in mind.

deepu105:
@sclausen take a look at this thread https://github.com/angular-translate/angular-translate/issues/849

sclausen:
@deepu105 I'm not sure, what you want to tell me with this thread. They discuss prefixing because of a very narrow corner case and don't come to a conclusion to implement this, so angular-translate still works like always and if I want to be compatible, I shouldn't implement prefixing.
Btw. the in the w3c document mentioned google translate service ignores this flag.

deepu105:
@sclausen what I'm trying to tell you is that the translate directive doesnt work when you use it in protractor coz protractor treats its as a html5 attribute and expects a boolean and not string as value. So I understand you want to maintain backword compatibility so what say you support both like below
```
@Component({
    selector: '[ng-translate],[translate]',
    ...
})
```
So that people who are not bothered by the issue can use `translate` and we (Jhipster folks, which is a huge user base btw, and others using this with protractor) can use `ng-translate`

sclausen:
@deepu105 that sounds really reasoned, I will add this. For the rest of my problems though I still have no solution.

deepu105:
@sclausen what do you think of the wrapper directive approach I posted above. It works without any problem for me. But im no ng2 expert so wouldn't know if it has any downfall compared to your original service ?

sclausen:
@deepu105 I don't think your wrapper would work for my desired featureset. The wrapper needs to have `<span translate="key"></span>` so it's not compatible to `angular-translate` anymore, since it was possible in `angular-translate` to use it also like this `<span translate>key</span>` and that `<span translate>{{'keys.' + variable}}</span>`.

gitnik:
So what's left to be done here? I'll be glad to help if some resources are needed

sclausen:
@gitnik I still have the problem, that 
`<p translate>hello_world</p>` doesn't update after first language change, but `<p translate [translate-values]="{value:'my-name'}">greetings</p>` works properly.
One strange thing I discovered is, that `translate.get` doesn't return after the first language change in the first example. Do you have any idea, why this happens?

gitnik:
I had a similiar issue. Maybe this can help you?
https://github.com/ocombe/ng2-translate/issues/288

src/translate.component.ts
```diff
@@ -0,0 +1,112 @@
...
+      this._elRef.nativeElement.innerHTML = res ? this.sanitizer.sanitize(SecurityContext.HTML, res) : this.key;
+      this._cdRef.markForCheck();
+    });
+  }
+
+  /**
+   * Clean any existing subscription to change events
+   * @private
+   */
+  _dispose(): void {
```
SamVerschueren: Add the `private` modifier and remove the `_`.

`private dispose(): void {`

src/translate.component.ts
```diff
@@ -0,0 +1,112 @@
...
+      });
+    }
+  }
+
+  /**
+   * updates the translation if the interpolation params change
+   * @param  changes
+   */
+  ngOnChanges(changes: { [key: string]: SimpleChange; }) {
+    if (changes["interpolationParams"] && this.key) {
```
SamVerschueren: single quotes -> `changes['interpolationParams']`

src/translate.component.ts
```diff
@@ -0,0 +1,112 @@
...
+  @Input('translate-values')
+  private interpolationParams: { [key: string]: string } = {};
+  private key: string;
+  onTranslationChange: EventEmitter<TranslationChangeEvent>;
+  onLangChange: EventEmitter<LangChangeEvent>;
+
+  constructor(
+    public sanitizer: Sanitizer,
+    public translate: TranslateService,
+    public _elRef: ElementRef,
```
SamVerschueren: Why the underscores?

src/translate.component.ts
```diff
@@ -0,0 +1,112 @@
...
+  TranslateService,
+  TranslationChangeEvent
+} from './translate.service';
+
+@Component({
+  selector: '[translate]',
+  template: '<ng-content></ng-content>'
+})
+export class TranslateComponent implements OnInit, OnChanges, OnDestroy {
+  @Input('translate')
```
SamVerschueren: It's bad practice to rename the bindings like this. Why not `@Input() translate: string` directly?

README.md
```diff
@@ -315,4 +315,3 @@ Ionic 2 is still using angular 2 RC4, but ng2-translate uses RC5. You should fix
 ## Additional Framework Support
 
 * [NativeScript](https://www.nativescript.org/) via [nativescript-ng2-translate](https://github.com/NathanWalker/nativescript-ng2-translate)
-
```
SamVerschueren: No unrelated changes

sclausen:
@SamVerschueren thanks for the remarks. I will fix this minor codestyle change requests along with fixing the faulty functionality. I'm currently a little out of spare time.

SamVerschueren:
No problem, we all have busy lives!

ocombe:
I have a question about this, let's say that you have the following html code:
```html
<div translate>
  Some text
  <span>some inner text</span>
</div>
```

then using the innerHTML as a key will receive: 
```

  Some text
  <span _ngcontent-bdn-1046="">some inner text</span>

```

Two problems here: the first is that we have the empty lines at the beginning and end which *can* be removed if you compress your html or use aot, and then it won't work anymore, but you probably won't use that as a key anyway because that'd be insane (so you'll use the `translate` attribute to specify a key).
Still, someone who uses:

```html
<div translate>
  Some text
</div>
```

might expect the key to be "Some text", and that will not be the case.

And the second problem is that angular adds attributes to inner elements, so that means that you can never use inner html elements in your keys because it won't work.
I won't even start with the innerHTML of elements that use bindings (like *ngIf), it adds comments all over the html...

This means that this directive will only work on really simple elements that only contain some text, that's still useful but I would love if it could do more. And to make sure that it works we should trim the content to remove leading/trailing spaces.

Do you think we could do something like that:
```html
<div translate>
  Some text
  <span translate>some inner text</span>
</div>
```

And in this case replace only the text at the first level, which would mean that here we have 2 translations: `Some text` and `some inner text` ? We replace those and ignore all the inner html?

sclausen:
@ocombe I think your example of
```
Some text
<span _ngcontent-bdn-1046="">some inner text</span>
```
as a key is a corner case which shoudln't be handled. Your translation keys are keys in a json file and if someone want's to have something like that as a translation key, I would be curious to hear the explanation why. In your third example, I agree, I would trim al the whitespace, like in angular-translate from @PascalPrecht ([translate.js#L167](https://github.com/angular-translate/angular-translate/blob/master/src/directive/translate.js#L167)), too.
I wanted to make this directive compatible to [angular-translate](https://github.com/angular-translate/angular-translate/blob/master/src/directive/translate.js) and ultimately support all its options, so something like this
``` html
<pre translate="TRANSLATION_ID"></pre>
<pre translate>TRANSLATION_ID</pre>
<pre translate translate-attr-title="TRANSLATION_ID"></pre>
<pre translate="{{translationId}}"></pre>
<pre translate>{{translationId}}</pre>
<pre translate="WITH_VALUES" translate-values="{value: 5}"></pre>
<pre translate translate-values="{value: 5}">WITH_VALUES</pre>
<pre translate="WITH_VALUES" translate-values="{{values}}"></pre>
<pre translate translate-values="{{values}}">WITH_VALUES</pre>
<pre translate translate-attr-title="WITH_VALUES" translate-values="{{values}}"></pre>
<pre translate="WITH_CAMEL_CASE_KEY" translate-value-camel-case-key="Hi"></pre>
```
@SamVerschueren that's also the reason I used `@Input('translate')`, because I didn't wanted to have translate to be the variable name of the key in some cases.

A lot has happened since I opened the pull request and much work has to be done to finish this directive/component. Currently my focus in angular2 is on other things, and I think I should rewrite this whole pr, but this may take a while.

ocombe:
No problem, I started rewriting it, using some of your work, but transforming it in a directive instead of a component :)

sclausen:
@ocombe That's great! :+1:

ocombe:
Let's keep this opened, I'll work from your PR so that you get the credits deserved, also this will help me not to forget to merge it :D

ocombe:
Done !
Thanks a lot to all of you and especially @sclausen for this, sorry it took so long !

sclausen:
So awesome! Thanks!

PR #44, opened 2016-02-24
Updated README.md to reflect correct path for prefix

nicohartto:
The README.md has a conflicting issue with path prefix, L#64 `var prefix = 'assets/i18n/'`

Since here we have slash again, between the two variables. This leads to // double slash problems in the URL. ```return this.http.get(`${this.sfLoaderParams.prefix}/${lang}${this.sfLoaderParams.suffix}`)``` 

[https://github.com/ocombe/ng2-translate/blob/master/src/translate.service.ts#L44](https://github.com/ocombe/ng2-translate/blob/master/src/translate.service.ts#L44)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,14 +45,14 @@ bootstrap(AppComponent, [
 })
 export class AppComponent {
     param: string = "world";
-    
+
     constructor(translate: TranslateService) {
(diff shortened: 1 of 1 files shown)


ocombe:
Thanks! Good catch

Issue #48, opened 2016-02-29
use interface instead of abstract class MissingTranslationHandler


ribizli:
It is much easier (more flexible) to implement an interface than extend a class in Typescript. ```MissingTranslationHandler``` has no added value being a an abstract class.

(I'd also check ```TranslateLoader``` for the same, but it has a play with DI, where class might be desirable)

ocombe:
You can implement an abstract class as well. How is it more flexible with an interface ?

ribizli:
In a class you can have only one superclass but more implemented
interfaces. It's a minor thing, anyway.

ocombe:
Hmm good to know, but can you write an interface that implements the abstract class and then implement this interface ?

ribizli:
I'd replace the abstract class with interface.

DanielSchuech:
I would agree in that case the the interface would be the better choice. I didn´t thought about this while implementing this feature.

Is your design maybe to complex? A class should only attend one responsibility to fulfill the Single Responsibility Principle. Maybe it would be better to devide your behabiour into different classes and MissingTranslationHandler inherit from just oneclass. That would be maybe better maintainable. But thats jsut an idee because I don`t know your use cases.

ribizli:
I've just stumbled into this thing when I was reading the source. For me the rule of thumb is to use interface if I don't need to deliver default functionality. Since the abstract class don't have any implemented feature, it could be cleaner to be a simple interface.

The test implementation shows also, that this abstract class is anyway used as interface only:
```typescript
class Missing implements MissingTranslationHandler
```

DanielSchuech:
@ribizli Yeah. As mentioned I agree with you.

ocombe:
Yup good points, I'll change that for the next release

ribizli:
I can prepare a PR for this and also create an NG2_TRANSLATE_PROVIDER for the default configuration. This is the proper angular2 way (like HTTP_PROVIDERS).

ocombe:
ah yes good idea, thanks

PR #50, opened 2016-03-01
configure of TranslateLoader and MissingTranslationHandler with angular2 DI only (#48)

ribizli:
Finally this PR has not much to do with #48, since Angular DI supports classes  and not interfaces.

Note! There are breaking changes in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,5 @@ With the given translation: `"HELLO": "hello {{value}}"`.
 - `interpolate(expr: string, params?: any): string`: Interpolates a string to replace parameters.
 
     `This is a {{ key }}` ==> `This is a value` with `params = { key: "value" }`
-- `flattenObject(target: Object): Object`:  Flattens an object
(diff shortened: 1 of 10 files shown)


ocombe:
Wow this looks this really nice, thanks a lot !
I will take a closer look at this asap (but maybe not until this week end).

ocombe:
I changed a few things, I renamed `NG_TRANSLATE_PROVIDERS` to `TRANSLATE_PROVIDERS` & I kept the methods `useLoader` & `setMissingTranslationHandler` because you might want to customize the service for one component only.

ribizli:
You can (should!) also do that via DI. I'm against the setter version. But
you decide.

ocombe:
Hmm how do you do that with DI in just one component?

ribizli:
In providers of component you can use the same provide method like in
bootstrap.

ocombe:
Oh nice, I didn't know that, I have to try :)
I'll remove the setters in the next version then.

PR #52, opened 2016-03-05
Introduce LangChangeEvent interface

sventschui:
Introduce LangChangeEvent interface

### DIFF
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -7,6 +7,11 @@ import 'rxjs/add/operator/map';
 
 import {Parser} from './translate.parser';
 
+export interface LangChangeEvent {
+    lang: string;
+    translations: any;
+}
+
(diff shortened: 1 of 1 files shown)


ocombe:
Hello, thanks for the PR.
Is there a reason for this change ?
Also can you add the type on line 69:
```ts
public onLangChange = new EventEmitter<LangChangeEvent>();
```

to
```ts
public onLangChange: EventEmitter<LangChangeEvent> = new EventEmitter<LangChangeEvent>();
```

sventschui:
It's easier to declare a event handler method (in users code) using `handle(event: LangChangeEvent)` than `handle(event: { lang: string, translations: any })`.

Adding the type is not necessary since ts will infer the type from the initial assignment.

ocombe:
Good points ! Thanks !
I'll merge it with a few other things for the next release.

PR #56, opened 2016-03-10
fix #53: dots in json keys is valid, but the parser doesn't acknowledge that

ribizli:

PR #62, opened 2016-03-17
fix#53 fixed the subkey issue.

Dinistro:
As i mentioned in #56 there is a problem with subkeys in strings:

```
'key' : 'value',
'key.subkey' : 'another value'
```

So here is the fix that does not harm all the other cases.


ocombe:
Fixed in https://github.com/ocombe/ng2-translate/releases/tag/v1.10.3. Thanks a lot for the PR.

PR #66, opened 2016-03-18
fix(pipe) - more robust when keys are falsy

yannickadam:
Translate pipe should not generate an error when receiving a falsy value such as undefined.


ocombe:
Fixed in https://github.com/ocombe/ng2-translate/releases/tag/v1.10.3

PR #76, opened 2016-03-29
Remove fromArray import

wmaurer:
`Observable.fromArray` does not exist in the latest beta of RxJS 5. It is deprecated in favour of `Observable.from`.
I've simply removed the import as it's not used in this file anyway.


ocombe:
Thanks !

wmaurer:
@ocombe no probs. FYI, this issue is causing my angular2 build with webpack to fail.

PR #83, opened 2016-04-04
update RxJS version to 4

jlast:
The RxJS 5 beta 2 bundle doesn't contain 'add/observable/of', which throws an error.
Beta 2 https://github.com/ReactiveX/rxjs/blob/9c887a49eda1a5ae87ae25b0f6eb49a9a63bf6f2/src/Rx.ts
Beta 4 https://github.com/ReactiveX/rxjs/blob/c4f6a5f8fc19fb2e340f0a4664dd91669923a400/src/Rx.ts

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "angular2": ">= 2.0.0-beta.10",
     "es6-shim": ">= 0.33.3",
     "reflect-metadata": ">= 0.1.2",
-    "rxjs": ">= 5.0.0-beta.2",
+    "rxjs": ">= 5.0.0-beta.4",
     "zone.js": ">= 0.6.4"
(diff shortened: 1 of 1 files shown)


ocombe:
Hello,

I'm sorry but I cannot force people to use RxJS beta 4 because the current Angular 2 beta uses RxJS beta 2 :-/

beagleknight:
Hi guys!

The angular rc.0 is out and it uses RxJS beta 5 :smile: 
See https://github.com/angular/angular/blob/master/CHANGELOG.md#200-rc0-2016-05-02 for more details.

ocombe:
Yep, looks like we can close this one :)

PR #92, opened 2016-04-15
fixed module naming in bundled ng2-translate.js

MartinWeindel:
At least in my environment (angular2 2.0.0-beta.14, rxjs 5.0.0-beta.2), the module naming in the bundled ng2-translate.js does only work after applying these changes.

### DIFF
--- a/bundles/ng2-translate.js
+++ b/bundles/ng2-translate.js
@@ -1,4 +1,4 @@
-System.registerDynamic("src/translate.pipe", ["angular2/core", "./translate.service", "angular2/src/facade/lang"], true, function($__require, exports, module) {
+System.registerDynamic("ng2-translate/src/translate.pipe", ["angular2/core", "ng2-translate/src/translate.service", "angular2/src/facade/lang"], true, function($__require, exports, module) {
   "use strict";
   ;
   var define,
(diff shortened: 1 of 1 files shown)


ocombe:
Hello,

the bundle file is generated before each commit, editing it won't fix the problem because it will be overwritten next time.
Are you including the bundle file as a script tag ?

MartinWeindel:
Hello,

sorry, I missed that the bundle is generated.
I have no experience with the systemjs-builder, but maybe its configuration is not quite correct?

And, yes the bundle file is included as a script tag in the index.html
`
  <script src='/assets/lib/angular2/bundles/angular2-polyfills.min.js'></script>
        <script src='/assets/lib/systemjs/dist/system.src.js'></script>
        <script src='/assets/lib/typescript/lib/typescript.js'></script>
        <script src='/assets/lib/rxjs/bundles/Rx.min.js'></script>
        <script src='/assets/lib/angular2/bundles/angular2.dev.js'></script>
        <script src='/assets/lib/angular2/bundles/router.dev.js'></script>
        <script src='/assets/lib/angular2/bundles/http.dev.js'></script>
        <script src='/assets/lib/ng2-translate/bundles/ng2-translate.js'></script>
`
The usage follows your description very closely (adding provider to bootstrap and configuration in application component).

Should I try to fix the generation and make another pull request?

ocombe:
I haven't tested the bundle in a while to be honest since I use webpack, let me check that before.

ocombe:
Unless you have some time, because I won't be able to take a look at it right now

PR #99, opened 2016-04-26
Update README.md

ozexpert:
with the latest ionic 2, the syntax should be link this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ bootstrap(AppComponent, [
 
 For Ionic 2 here is a complete bootstrap with configuration:
 ```ts
+import {Http} from 'angular2/http';
 import {provide} from 'angular2/core';
 import {TranslateService, TranslateLoader, TranslateStaticLoader} from 'ng2-translate/ng2-translate';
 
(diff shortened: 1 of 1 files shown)


mhartington:
Hi, Mike from ionic here!

So there seems to be some confusion. 
The example here is more for typescript, so the annotation in `useFactory: (http: Http)` are ok.

Also, ` deps: [[Http]]` is not correct either. You're probably getting that from the `get parameters()` in a JS ionic2 app. This wouldn't be the case here since the example is in Typescript.

mhartington:
I can confirm that if you're using ES6 and not Typescript, you can simply use 

```
@App({
  template: '<ion-nav [root]="rootPage"></ion-nav>',
  config: {}, // http://ionicframework.com/docs/v2/api/config/Config/
    providers: [
    provide(TranslateLoader, {
      useFactory: (http) => new TranslateStaticLoader(http, 'assets/i18n', '.json'),
      deps: [Http]
    }),
    TranslateService
  ]
})
```

It's the same code, just without the annotation on `http`

ocombe:
Thanks mike ! :)

ozexpert:
Thanks for clarification @mhartington

ocombe:
So @ozexpert can you update your PR so that I can merge it ? Just commit the update to the same branch and it should update the PR.

PR #102, opened 2016-04-26
fix(Parser): accept falsy values

nsanitate:
Parser should interpolate with falsy values like empty strings and zero.


ocombe:
Thanks a lot !

PR #140, opened 2016-06-02
[Issue#87] fix target undefined for missing translation

brosenberger:
when the keys are shifted, somewhen the target gets set to undefined (L39). the loop terminates only after the next interation. in the meanwhile an exception is thrown and the further translation is aborted -> therefore added a check for undefined to exit the loop correctly without exception

### DIFF
--- a/src/translate.parser.ts
+++ b/src/translate.parser.ts
@@ -32,7 +32,7 @@ export class Parser {
         key = '';
         do {
             key += keys.shift();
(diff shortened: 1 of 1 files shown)


DavidWiesner:
I've created an build with this pull request to try this out run
`npm install "https://github.com/DavidWiesner/ng2-translate/tarball/808d27c5bd66428d58c79aea61bcd7b63dcc0b26"`

Issue #146, opened 2016-06-12
Remove Http dependency from the TranslateService class


dscheerens:
It appears the `TranslateService` requires an `Http` instance when constructed. This dependency, however, does not seem to be used anywhere in the `TranslateService` (maybe a leftover from earlier days?). Removing it could be easier for testing, when you are not using the `TranslateStaticLoader` (which does have a dependency on `Http`). That way there is no need to load any of the Http related providers, which keeps your tests cleaner!

Anyway, thanks for your nice work on this project!

yoava:
+1 
I'm using a custom TranslateLoader and have no need for Http, this dependency just bloats my served js.

ocombe:
Yes I need to distribute the loaders separately

dscheerens:
That is probably also a good idea, but what I was actually referring to is the `Http` dependency in the constructor of the `TranslateService` class.

ocombe:
yes, but if I externalize the loaders, this will go away

yoava:
I guess :)
But removing the ```Http``` dependency will only take 1 minute :) (it's not really used)
Here's a pull request with the fix - https://github.com/ocombe/ng2-translate/pull/149

PR #149, opened 2016-06-19
#146 - Remove Http dependency from the TranslateService class

yoava:
Removed the unused ```Http``` dependency.


ocombe:
Thanks ! Could you update the readme as well please ?

PR #152, opened 2016-06-20
Remove deprecated @App decorator

eddhannay:
- Update Readme to remove [deprecated @App usage](https://github.com/driftyco/ionic/blob/2.0/CHANGELOG.md#200-beta8-2016-06-06)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,8 @@ For Ionic 2 here is a complete bootstrap with configuration:
 ```ts
 import {TranslateService, TranslateLoader, TranslateStaticLoader} from 'ng2-translate/ng2-translate';
 
-@App({
+@Component({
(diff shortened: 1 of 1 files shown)


eddhannay:
I missed this in the last Readme update, sorry for the multiple PRs!

ocombe:
ionic 2 uses @Component instead of @App now ?

eddhannay:
Yeah, it was change in [https://github.com/driftyco/ionic/blob/2.0/CHANGELOG.md#steps-to-upgrade-to-beta-8](beta 8). Also the `config` property has moved (as per that upgrade guide), so I'll move that out too   - it's unused in the example code anyway!

ocombe:
That's a good choice that they did here! It was confusing

PR #162, opened 2016-07-04
typo

mbakker96:
Added '(' by onLangChange


ocombe:
Good catch, thanks :)

PR #170, opened 2016-07-14
fix(TranslatePipe): update translations after translateService.set called

mrkosima:
It would be great automatically update translation when `translateService.set` was called


ocombe:
This is really good work, and a good idea, thanks

ocombe:
Merged in 2.4

PR #173, opened 2016-07-19
Getter and setter for array of available languages.

h3xed:
Getter and setter methods for provide array of available languages.

For example:
On translate library init i can declare languages available in system by call:
`translate.setAvailableLangs(['en', 'pl', 'de']);`
And next i can render menu by get available (but not necessary currently loaded) languages:
`this.availableLangs = translate.getAvailableLangs();`


szneba:
+1

ocombe:
This makes sense, good idea, thanks !

ocombe:
Merged in 2.4

PR #197, opened 2016-08-16
Add typings to package.json

wmaurer:
"typings" key is required in package.json so that tsc knows where to find the typings for this module


ocombe:
Good call thanks !

PR #205, opened 2016-08-18
Fixing which language to use

jpinkster:
Thank you so much for adding the the `TranslateService.getBrowserLang()` .. this is awesome. I just wanted to update how you are determining which language to use when there are more than 1 set in the browser. 

If the browsers languages have more than one language, just grab the users number one preference. For instance, if I am a spanish user in the US and my device language is in spanish, `window.navigator.language` shows 'es' and my `window.navigator.languages = ['es', 'en-US', 'en']` .. just grab `'es'`

### DIFF
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -355,7 +355,7 @@ export class TranslateService {
         }
         let browserLang: any;
         if (typeof window.navigator['languages'] !== 'undefined' && window.navigator['languages'].length > 0) {
-            browserLang = window.navigator['languages'][0].indexOf('-') !== -1 || window.navigator['languages'].length < 2 ? window.navigator['languages'][0] : window.navigator['languages'][1];
+            browserLang = window.navigator['languages'][0];
         } else {
             browserLang = window.navigator['language'] || window.navigator['browserLanguage'];
         }


ocombe:
While this is usually true, all browser behave differently.
I had to add this check for some version of IE or safari (not sure which one) :)

jpinkster:
Haha yeah man .. probably an IE issue. Just trying to figure out how you were able to determine the current logic makes the most sense. I have updated my logic and tested across all browsers (chrome, safari, firefox, and IE10+) and it is pulling the users default language in perfectly. If you keep your existing logic, it is mistakenly taking the users secondary language, which is inaccurate. 

To truly encompass all browser support the logic should be: 
```
public getBrowserLang(): string {
    if (typeof window === 'undefined' || typeof window.navigator === 'undefined') {
        return undefined;
    }
    let browserLang: any;
    
    browserLang = window.navigator.languages ? window.navigator.languages[0] : null;
    browserLang = browserLang || window.navigator.language || window.navigator.browserLanguage || window.navigator.userLanguage;
    if (browserLang.indexOf('-') !== -1)
        browserLang = browserLang.split('-')[0];

    if (browserLang.indexOf('_') !== -1)
        browserLang = browserLang.split('_')[0];
    
    return browserLang;
}
```

And I could really argue that it should be:
```
public getBrowserLang(): string {
    if (typeof window === 'undefined' || typeof window.navigator === 'undefined') {
        return undefined;
    }
    let browserLang: any;
    
    browserLang = window.navigator.languages ? window.navigator.languages[0] : null;
    browserLang = browserLang || window.navigator.language || window.navigator.browserLanguage || window.navigator.userLanguage;
   
    return browserLang;
}
```
.. omit the stripping of the location part of the locale, and let the developer utilizing this method determine if they want to strip that off or not... again this is a helper method for us to determine what the users current default language is set to so that we can default to their true setting.

Please reconsider merging this change.

sean-perkins:
@ocombe 

While browsers may have different ways to organize languages, it's unlikely that pulling any other index, other than the primary index, would make sense to the user (based on their preferences). 

`window.navigator.language` should equal the same as `window.navigator.languages[0]`. 

Since this is a helper function, it either should return the preferred user's language choice, or have optional variables to select the specific language index you may want, to alleviate your concern about some version of IE or Safari that caused the initial logic behind the current function.

PR #229, opened 2016-09-07
fix(TranslateModule): use exported factory function to enable AOT compiling

frederikschubert:
I haven't tested it with this module but in other cases this change fixed the problem.

Should close https://github.com/ocombe/ng2-translate/issues/218

### DIFF
--- a/ng2-translate.ts
+++ b/ng2-translate.ts
@@ -13,6 +13,10 @@ export default {
     providers: [TranslateService]
 };
 
+export function translateLoaderFactory(http: Http) {
+    return new TranslateStaticLoader(http);
+}
+
 @NgModule({
     imports: [HttpModule],
     declarations: [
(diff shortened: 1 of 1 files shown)


ocombe:
Ah thanks, that's a weird solution, I'll take a look asap to know if it works

frederikschubert:
Yes it is weird but others have found similar [problems/solutions](https://github.com/ngrx/store/pull/208#issuecomment-245598623) for the AOT compiler.

BrainsoftLtd:
Please merge this, because other projects are running into the same problems (angular2-seed-advanced for one)...

ocombe:
Yes sorry about that, I'll merge this as I don't see how it can do any harm anyway :)
Thanks for the PR @frederikschubert

BrainsoftLtd:
Wow, that was fast :)

ocombe:
You can try version 2.4.4 :-)

PR #254, opened 2016-10-03
Update translate.service.ts

MadDeveloper:
Allows to avoid this case (used in the [plunker](http://plnkr.co/edit/btpW3l0jr5beJVjohy1Q?p=preview)):

```
translate.addLangs(["en", "fr"]);
translate.setDefaultLang('en');

let browserLang = translate.getBrowserLang();
translate.use(browserLang.match(/en|fr/) ? browserLang : 'en');
```

with modification:

```
translate.use(browserLang.match(/en|fr/) ? browserLang : translate.getDefaultLang());
```

### DIFF
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -104,6 +104,10 @@ export class TranslateService {
         this.defaultLang = lang;
     }
 
+    public getDefaultLang(): string {
+        return this.defaultLang;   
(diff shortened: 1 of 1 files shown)


ocombe:
This makes sense :) I'll merge it at the next release cycle

PR #261, opened 2016-10-04
fix(translate): treat empty string as valid translation - fixes #231

### LINKED ISSUE #231: Treat empty string as valid translations (at least conditionally)
**I'm submitting a ...**
```
[x] bug report
```

**Current behavior**
If translation yields empty string, the translation is considered missing.

**Expected/desired behavior**
Empty strings, should either be considered valid translations **always**, or with a conditional switch, like
```typescript
translateService.tolerateEmptyStrings();
```

See [this plunker](http://plnkr.co/edit/J9RscKq79uOqkx3UkPs5?p=preview)

Use case is pretty much described in #223 . When localizing links, this is essential. 

** Comments **
One could add a space there:
```json
{
  "SOME_KEY": " "
}
```
but this is error-prone as such kinda redundant whitespace can be lost when localization team does its job.

* **ng2-translate version:** 2.4.3

* **Angular version:** 2.0.0-rc.6
 
* **Browser:** all
 
* **Language:** all

SamVerschueren:
This fix fixes #231. Not sure if I should add the `bundles/ng2-translate.js` to the PR though. Let me know if I should remove it.

The thing is that an empty string in JavaScript is `false` and thus `'' || 'foo'` will result in `foo`.


ocombe:
Thanks !

SamVerschueren:
Thanks for merging!

PR #263, opened 2016-10-04
refactor(barrel): add barrel - fixes #253

### LINKED ISSUE #253: Why not `ng2-translate/index`?
Hi @ocombe. I was just wondering, why do we have to import from `ng2-translate/ng2-translate`? Why can't we change the main entrypoint to `index.ts`? I believe webpack automatically handles barrels so you can `import { TranslateService } from 'ng2-translate';` and SystemJS is easily configurable. Maybe it will support barrels in the future as well.

Maybe it's done on purpose and it has a valid reason, but it feels weird for me to type `ng2-translate` twice :).

SamVerschueren:
Not sure if this is enough to get everything working. Somehow, when I used `npm link` here locally, it failed with AoT because it said it couldn't find `index.d.ts` and the source file `index.ts` which is odd because they are 100% sure present. Not sure how I can test the AoT stuff?


ocombe:
Yeah npm link doesn't really work with typescript libs :(
If you try it before your update, it should still fail right ?

ocombe:
Can you try with
```ts
resolve: { 
     modules: [ path.join(__dirname, "node_modules") ] 
}
```
in your webpack config, apparently it fixes the npm link issue

SamVerschueren:
I use SystemJS so not really an idea how I should test it with Webpack though :p. I can always try to run the example here in this repo.

SamVerschueren:
Getting the exact same error when I try to use `npm link` with `master` indicating that it can't find `ng2-translate.d.ts`.

ocombe:
Ok "good", then it's not a problem with your changes

SamVerschueren:
I'm now testing it with the webpack example from this repo.

ocombe:
yeah I think the webpack example is outdated with the last changes from rob

SamVerschueren:
Getting an error when I try to run the example directly (not even without `npm link`). We should look into updating these examples to the latest Angular and ng2-translate.

For now, it seems that the only way we can test this is by releasing a new version and see what happens. I think not much can go wrong as the tests are still working and it still support `ng2-translate/ng2-translate` imports. I know it's playing with fire, but it's sad to say, do we have another choice? :s

SamVerschueren:
> For now, it seems that the only way we can test this is by releasing a new version and see what happens.

Just realised you could also release a beta version first and then we can test with that one.

SamVerschueren:
Friendly ping :). No rush if you don't have the time to look into it.

PR #277, opened 2016-10-12
Support single quotes in param values

### LINKED ISSUE #276: Support single quotes for values in translate params
```
[x] feature request
```

**Current behavior**
We are migrating from angular-i18n to ng2-translate
When we use `<div [innerHTML]="'HELLO' | translate:{ param: 'value'}"></div>` it throws an error from https://github.com/ocombe/ng2-translate/blob/master/src/translate.pipe.ts#L98

**Expected/desired behavior**

Since its not possible to pass double quotes from inside directives like above it would be nice if single quotes can be supported and handled similar to https://github.com/ocombe/ng2-translate/blob/master/src/translate.pipe.ts#L96
Currently the only way to do this is to use  `<div [innerHTML]="'HELLO' | translate:{ param: &quote;value&quote;}"></div>` which is not very nice for readability

Use the pipe from within directives

* **ng2-translate version:** latest

* **Angular version:** 2.0.0 final
 
* **Browser:** all
 
* **Language:** TypeScript

deepu105:
Add support for single quotes in param values
Remove redundant replace of proper keys
Add more test cases

fix #276


deepu105:
when is your next release? also do you think the directive #31 will progress? we would be very interested to have it

ocombe:
The release cycle is automatic, when a PR is merged and the tests pass it is automatically released, you can try it with version 3.1.1
About the directive it will be merged, but I need to find some free time to work on it first, so no ETA yet

PR #283, opened 2016-10-19
feat: declension support added

pawwach:
I've added a declension support - you can see examples in unit tests. If you have any questions I am at your disposal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ In the html page:
 - `reloadLang(lang: string): Observable<string|Object>`: Calls resetLang and retrieves the translations object for the current loader
 - `resetLang(lang: string)`: Removes the current translations for this lang. /!\ You will have to call `use`, `reloadLang` or `getTranslation` again to be able to get translations
 - `getBrowserLang(): string | undefined`: Returns the current browser lang if available, or undefined otherwise 
-- `setDeclensionRule(lang: string, declensionRule: Function)`: Sets a declension rule for a given language
+- `setPluralizationRule(lang: string, pluralizationRule: Function)`: Sets a pluralization rule for a given language
 
 #### Write & use your own loader
 If you want to write your own loader, you need to create a class that implements `TranslateLoader`.
@@ -249,10 +249,10 @@ export class SharedModule {
 }
 ```
 
-#### How to support a declension
-Most languages require the proper word declension. If you want to support it in a particular language, you have to define declension rules when the `TranslateService` is being initialized:
+#### How to support a pluralization
+Most languages require the proper word pluralization. If you want to support it in a particular language, you have to define pluralization rules when the `TranslateService` is being initialized:
 ```ts
-translate.setDeclensionRule('en', (value: any) => {
+translate.setPluralizationRule('en', (value: any) => {
     return value === 1 ? 0 : 1;
 });
 ```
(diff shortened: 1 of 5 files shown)


ocombe:
Wow that's a very nice PR thanks !
I feel like we should just call it plural or pluralize, I had to search on google to find what "declension" means :)

pawelwachala:
You,re right, I changed "declension" to "pluralization".

aitboudad:
I want to use https://github.com/messageformat/messageformat.js for pluralization, is that possible?

pawwach:
@aitboudad No, unfortunately not. It's my own solution, pretty simple. But your question inspired me to introduce one more improvement: named pluralization rules. In current solution, you can use one pluralization rule per one language. If named pluralization rules will be added, you will be able to define another rule for pluralization and another for handling ordinal numbers. For example:

Rules still will have to be defined in code (when ng2-translate is initialized). I think that containing their in JSON files is bad idea.

I have {{value}} cat{{@value(|s)}} => I have 1 cat, I have 2 cats etc. (default pluralization rule for English)
My {{"ruleName"@value(st|nd|rd|th)}} cat is very cute => i.e. My 2nd cat is very cute (named pluralization rule "ruleName" for English)

Of course syntax is negotiable issue.

pawwach:
@aitboudad No, unfortunately not. It's my own solution, pretty simple. But your question inspired me to introduce one more improvement: named pluralization rules. In current solution, you can use one pluralization rule per one language. If named pluralization rules will be added, you will be able to define another rule for pluralization and another for handling ordinal numbers. For example:

`I have {{value}} cat{{@value(|s)}}` => `I have 1 cat`, `I have 2 cats` etc. (default pluralization rule for English)
`My {{"ruleName"@value(st|nd|rd|th)}} cat is very cute` => i.e.` My 2nd cat is very cute` (named pluralization rule "ruleName" for English)

Rules still will have to be defined in code (when ng2-translate is initialized). I think that containing their in JSON files is bad idea.

Of course syntax is negotiable issue.

pawwach:
@aitboudad No, unfortunately not. It's my own solution, pretty simple. But your question inspired me to introduce one more improvement: named pluralization rules. In current solution, you can use one pluralization rule per one language. If named pluralization rules will be added, you will be able to define another rule for pluralization and another for handling ordinal numbers. For example:

`I have {{value}} cat{{@value(|s)}}` => `I have 1 cat`, `I have 2 cats` etc. (default pluralization rule for English)
`My {{"ruleName"@value(st|nd|rd|th)}} cat is very cute` => i.e. `My 2nd cat is very cute` (named pluralization rule "ruleName" for English)

Rules still will have to be defined in code (when ng2-translate is initialized). I think that placing them in JSON files is bad idea.

Of course syntax is negotiable issue.

DethAriel:
This looks redundant to me - Angular 2 already has i18nPlural pipe, which deals with pluralization, handles locale-specific pluralization categories, and is generally more functional. Plus, it can already be used with `ng2-translate` like [in this plunker](http://plnkr.co/edit/OcRw07B2Ow1wzU0epAPm?p=preview). My vote would be to simpify that scenario in favor of reinventing the wheel.

E.g. one should not have to do this:
```ts
@Component({
    selector: 'my-inbox',
    template: `
    <div *ngIf="pluralizationMap">
      {{ inboxMsgs.length | i18nPlural:pluralizationMap }}
    </div>`,
})
export class InboxComponent implements OnInit {
    inboxMsgs = [];    
    pluralizationMap = null;
    
    constructor(private translate: TranslateService) {}
    
    ngOnInit() {
      this.translate.get('INBOX_MESSAGES_PLURAL')
        .subscribe(map => this.pluralizationMap = map);
    }
}
```

but instead have `ng2-translate` do the job internally, something like
```ts
@Component({
  selector: 'my-inbox',
  template: `
    {{ inboxMsgs.length | translatePlural:'INBOX_MESSAGES_PLURAL' }}
    OR
    {{ inboxMsgs.length | translatePlural:{ pluralRulesKey: 'INBOX_MESSAGES_PLURAL', otherParams: params } }}
  `
})
export class InboxComponent {
  inboxMsgs = [];
  params: { some: 'param' };
}
```

tommueller:
Is the pluralization done by now?

Issue #258, opened 2016-10-03
[3.0.0] Webpack example not working 


rbaumi:
**I'm submitting a ...**  (check one with "x")

```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**

> Uncaught TypeError: Cannot read property 'prototype' of undefined

**Expected/desired behavior**
to work out of the box

**Reproduction of the problem**
git clone https://github.com/ocombe/ng2-translate.git
cd ng2-translate/examples/webpack
npm install
npm start
watch console output...

**ng2-translate version:** 3.0.0

**Browser:** [all] 
don't think it's browser related

ocombe:
What repository are you cloning ?

julb:
Maybe due to the fact that the library isn't distributed in commonjs format ?

ocombe:
yes I need to update the examples with the new umd format

ocombe:
This has been updated, closing it now.

Issue #273, opened 2016-10-07
Issue with webpack statically extracted dependencies


gamazeps:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**
Throwing a warning that is turned into a hard error with static files build.

```
WARNING in ./~/ng2-translate/bundles/ng2-translate.js
Critical dependencies:
1:2208-2215 require function is used in a way in which dependencies cannot be statically extracted
1:2254-2261 require function is used in a way in which dependencies cannot be statically extracted
 @ ./~/ng2-translate/bundles/ng2-translate.js 1:2208-2215 1:2254-2261
```

Which leads to this output in the console with static files:
```
Uncaught TypeError: Cannot read property 'resolve' of undefined
Uncaught TypeError: Cannot read property 'forRoot' of undefined
```

**Reproduction of the problem**

```sh
git clone git@github.com:angular/angular2-seed.git
cd angular2-seed
npm install
npm install --save ng2-translate
```

In `src/vendor.browser.ts`

```typescript
import 'ng2-translate';
```

In `src/app/app.module.ts`

```typescript
import { TranslateModule } from 'ng2-translate'; // <<<<<<<<<<<<<<<<<<<< <  added

@NgModule({
  declarations: [AppComponent, About, RepoBrowser, RepoList, RepoDetail, Home],
  imports     : [
    BrowserModule,
    FormsModule,
    HttpModule,
    RouterModule.forRoot(rootRouterConfig),
    TranslateModule.forRoot() // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< <added
  ],
  providers   : [Github, {provide: LocationStrategy, useClass: HashLocationStrategy}],
  bootstrap   : [AppComponent]
})
export class AppModule {

}
```

**Please tell us about your environment:**

* **ng2-translate version:** 3.1.0

* **Angular version:** 2.0.0
 
* **Browser:** [Chrome 53]

ocombe:
Can you change `import { TranslateModule } from 'ng2-translate'` to `import { TranslateModule } from 'ng2-translate/ng2-translate'` and see if it works?

gamazeps:
Tried it and it doesn't work infortunately.

This gives a hard error:

```
ng2-translate.js:1Uncaught SyntaxError: Unexpected token import
[...]

PR #285, opened 2016-10-21
Webpack example with ng2-translate 3.1.2

fgladisch:
Hi,

this PR attempts to fix issues #258 and #273 with an updated webpack example that includes...
- ng2-translate 3.1.2
- Angular 2.1.1
- The fix (or dirty workaround?!) discussed in #273: loading ng2-translate with Babel
- Minor style changes
- Downgrade to a stable webpack version

Although this example works, I could not get rid of this warning:

```
WARNING in ./~/ng2-translate/bundles/ng2-translate.js
Critical dependencies:
58:100-107 require function is used in a way in which dependencies cannot be statically extracted
58:152-159 require function is used in a way in which dependencies cannot be statically extracted
 @ ./~/ng2-translate/bundles/ng2-translate.js 58:100-107 58:152-159
```

Anything missing?

### DIFF
--- a/examples/webpack/package.json
+++ b/examples/webpack/package.json
@@ -7,30 +7,32 @@
     "start": "webpack-dev-server --inline --progress --port 3000"
   },
   "dependencies": {
-    "@angular/common": "2.0.0-rc.5",
-    "@angular/compiler": "^2.0.0-rc.5",
-    "@angular/core": "2.0.0-rc.5",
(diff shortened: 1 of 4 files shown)


ocombe:
Hello, and thank you for the PR, but I've updated the example with the latest release and it doesn't need to use babel to work, so that's better I think.

Issue #293, opened 2016-10-31
Added support for Language Culture Name


thorsten:
**I'm submitting a ...**  
```
[ ] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[x] feature request
```

**Current behavior**
getBrowserLang() returns just the language code, e.g. "de" or "en". In our current project there's the need to use the language culture name, e.g. "de-DE" or "en-US".

**Expected/desired behavior**
Add a new function getBrowserCultureLang(), that returns the language culture name, e.g. "de-DE"

**What is the motivation / use case for changing the behavior?**
The need in a project for this feature.

**Please tell us about your environment:**

* **ng2-translate version:** 3.1.0

* **Angular version:** 2.1.0
 
* **Browser:** all
 
* **Language:** all

chpasha:
the methods getBrowserLang and getBrowserCultureLang have 99% same code or do I miss anything? getBrowserLang should actually reuse getBrowserCultureLang and split result with - or _

thorsten:
Yes, you are right. Good hint. @ocombe, should I add another PR with this improvement?

ocombe:
yes please :)

PR #295, opened 2016-11-01
Added support for Language Culture Name

thorsten:
Patch for #293

### DIFF
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -361,6 +361,11 @@ export class TranslateService {
         this.translations[lang] = undefined;
     }
 
+    /**
+     * Returns the language code name from the browser, e.g. "de"
+     *
+     * @returns {any}
+     */
     public getBrowserLang(): string {
         if(typeof window === 'undefined' || typeof window.navigator === 'undefined') {
             return undefined;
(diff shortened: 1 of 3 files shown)


ocombe:
Thanks, could you also update the readme to add the description of this new method ?

thorsten:
Yes, sure! Will update the documentation asap.

src/translate.service.ts
```diff
@@ -361,6 +361,11 @@ export class TranslateService {
         this.translations[lang] = undefined;
     }
 
+    /**
+     * Returns the language code name from the browser, e.g. "de"
+     *
+     * @returns {any}
```
ocombe: change `@returns {any}` to `@returns string`

src/translate.service.ts
```diff
@@ -379,4 +384,20 @@ export class TranslateService {
 
         return browserLang;
     }
+
+    /**
+     * Returns the culture language code name from the browser, e.g. "de-DE"
+     *
+     * @returns {any}
```
ocombe: change `@returns {any}` to `@returns string`

src/translate.service.ts
thorsten: ack

src/translate.service.ts
thorsten: ack

ocombe:
Perfect thanks

PR #296, opened 2016-11-01
docs: update on how to use with NativeScript

NathanWalker:
The plugin is no longer needed since the feature it provided is now available in core. Just updates the note to point to a comment providing help on how to set it up.


ocombe:
Thanks !

PR #299, opened 2016-11-02
fix systemjs example

SamVerschueren:
Upgraded the SystemJS example. Tried to upgrade the Webpack example as well but received an error, no idea why though but I'm not a Webpack user so hard to troubleshoot :).


examples/systemjs/src/app.component.ts
```diff
@@ -15,12 +15,14 @@ import {TranslateService} from 'ng2-translate/ng2-translate';
...
   `,
 })
-export class AppComponent {
-    constructor(private translate: TranslateService) {
-        translate.addLangs(["en", "fr"]);
-        translate.setDefaultLang('en');
+export class AppComponent implements OnInit {
+    constructor(private translate: TranslateService) { }
+  
+    ngOnInit() {
```
SamVerschueren: Used `ngOnInit` because it's best practice to keep the constructor as light as possible.

ocombe:
Could you update it to 4.0.0? the new main file is `bundles/ng2-translate.umd.js` (the index.js still exists for compatibility)

SamVerschueren:
Done!

ocombe:
thanks !

SamVerschueren:
Cool! Excited that all these things are being improved step by step :).

PR #300, opened 2016-11-02
Improvement for #293

thorsten:
As requested.


src/translate.service.ts
```diff
@@ -396,13 +396,12 @@ export class TranslateService {
      * @returns string
      */
     public getBrowserLang(): string {
-        if(typeof window === 'undefined' || typeof window.navigator === 'undefined') {
+        let browserLang = this.getBrowserCultureLang();
+
+        if(browserLang === 'undefined') {
```
SamVerschueren: You mean `browserLang === undefined`?

ocombe: I'd rather keep `typeof window === 'undefined'` because undefined can be defined (yeah, I know :D)

SamVerschueren: I was referring to the browserLang check, it doesn't use `typeof`. So probably it should be changed to

```js
if(typeof browserLang === 'undefined') {
```

thorsten: You're right, it would be more consistent.

PR #311, opened 2016-11-06
Fix TranslateService.get when called while pending

### LINKED ISSUE #308: translate.get(...).flatMap(...) not working in pending state
```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[x] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```
Not sure if it's a bug or misuse by me...

**Current behavior**
Using `translate.get('...').flatMap(...).subscribe(observer)` does not work if `this.pending` is set.

**Expected/desired behavior**
The `observer` gets called with the translation.

Made a small [plunker](http://plnkr.co/edit/smNbPyTIp8LCyVOe0YGz?p=preview). Calling `doIt()` from `ngOnInit()` does not work, calling it afterwards works just fine. Although I didn't dig too deep it looks like to be caused by `this.pending`.

imgx64:
Fixes #308

### DIFF
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -6,6 +6,7 @@ import "rxjs/add/observable/of";
 import "rxjs/add/operator/share";
 import "rxjs/add/operator/map";
 import "rxjs/add/operator/merge";
+import "rxjs/add/operator/mergeMap";
 import "rxjs/add/operator/toArray";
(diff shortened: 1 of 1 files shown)


imgx64:
#312 is better

PR #318, opened 2016-11-10
Update README.md

Mathou54:
Add documentation for translateService.get(). It return an observable of key if the key was not found (= already translated), which is very useful when developing a component.


ocombe:
Thanks !

PR #333, opened 2016-11-23
Readme fixes for examples using pipe values

pndewit:
The pipe examples using values gave me errors. Maybe it is an idea to also show an example of a regular string as value?
e.g.
<div>{{ 'HELLO' | translate:{value: "Some value"} }}</div>

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ import {TranslateService} from 'ng2-translate';
 @Component({
     selector: 'app',
     template: `
-        <div>{{ 'HELLO' | translate:{value: param} }}</div>
+        <div>{{ 'HELLO' | translate:'{value: "' + param + '"}' }}</div>
     `
 })
 export class AppComponent {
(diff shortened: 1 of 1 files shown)


ocombe:
Hmm the example is supposed to work, if that's not the case then it's a bug.
And yes we could add a new simpler example with a string value

pndewit:
http://plnkr.co/edit/mwE3PCTo7PTlMgH7ljIQ?p=preview
See my Plunker.

ocombe:
Damn it you're right, here is how it should be: http://plnkr.co/edit/yJPzPnuTQ0qnrNCPIjbU?p=preview
I wonder why you can't use the object directly in the template

pndewit:
Ahhh! This works for me though, thanks!
Haven't checked out your code yet, will do so soon!

Love your app, thanks for all the effort you put into this!

ocombe:
Ok I've updated the readme accordingly, thanks a lot for reporting this !

Issue #332, opened 2016-11-23
setDefaultLang is not working as fallback language


pndewit:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**
I use the following two calls to set the current language to French and the default (fallback) language to English.

```
TranslateService.setDefaultLang('en');
TranslateService.use('fr');
```

So if a translation is missing from the French translations file, it should fall back to the English translations file. This is not the case though.

The use function successfully adds the French translations, but the setDefaultLang is not adding the English translations. The English translations should even be the default (making the setDefaultLang call even unnecessary). I have also tried to call the addLangs function, but this is not working either.

**Expected/desired behavior**
I would like ng2-translate to load in all the translation files inside the i18n directory automatically to so that I wouldn't even have to use the use function and let the browser of the client decide which translations to try and use (falling back to the fallback language if it is not available).

I would expect that the "en.json" to be injected by default (unless specified otherwise by setDefaultLang) and I would also expect that the setDefaultLang to inject the given translations file.

Lastly (but I am not sure if this is according to what it is supposed to do) I would expect the addLangs function to inject the given translations. But I don't think that a list of available languages is something that ng2-translate 'should' be tracking anyway.

**Reproduction of the problem**
If the current behavior is a bug or you can illustrate your feature request better with an example, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:btpW3l0jr5beJVjohy1Q).

http://plnkr.co/edit/4M7kUf0H8utxrF7bNskL?p=preview

**What is the motivation / use case for changing the behavior?**
The fallback language does not work unless the use function is called for it.

**Please tell us about your environment:**
Angular 2 v2.2.1 (also tried v2.2.2) with Webpack. But it is reproducible in other environments (check the Plunker linked above using your given starting point).

* **ng2-translate version:**
latest (v4.0.1)

* **Angular version:**
v2.2.1 / v2.2.2

* **Browser:**
Chrome
 
* **Language:**
TS

pndewit:
I could help with a PR if you like!

If anyone faces the same issue. I simple workaround is doing a use for the fallback language before setting it to the initial language (e.g. browser's language):
`
TranslateService.setDefaultLang('en');
[...]

PR #336, opened 2016-11-24
#332: Added default fallback language to English

pndewit:
#332 Default fallback language was never set and setting a default language did not request for the translations file if the language was not recognised.

Also added a version to the package.json since: 

> The most important things in your package.json are the name and version fields. Those are actually required, and your package won't install without them.

Source: https://docs.npmjs.com/files/package.json#version

Without this version I could not reference to my fork version on GitHub from my own project's package.json file. If (for some reason) you do not want it, you may remove it of course.

Let me know what you think!


ocombe:
The version is automatically generated on Travis when I push / merge commits (based on the commits messages), that's why it's empty in package.json :)

pndewit:
@ocombe Ahhh right, my bad! Shall I revert the commit?

SamVerschueren:
Just remove it and push the changes. We can squash commits here.

pndewit:
Done. :)

ocombe:
So my point of opinion on this PR:
- We should not set the default language to "en" automatically, we just don't know how people want to name it, and maybe for some reason they don't want a default language
- Is it really necessary to have a `DefaultLangChangeEvent` ? I don't see any use case, you define the default lang at bootstrap and then you forget about it, it's not really something that you should change
- If we don't use the `DefaultLangChangeEvent`, then `changeDefaultLang` becomes unecessary as well

ocombe:
So in the end, the only thing that you should do in this PR is: get the default lang file when the current translation is missing (if the default lang is missing too)

ocombe:
And sorry if I seem a bit blunt here, I'm just making sure that we don't bloat the code too much by adding stuff that isn't necessary. But rest assured that I appreciate the time that you took to do this PR, and it's a good one (clean code and tests!)

pndewit:
@ocombe Thank you for your feedback! No worries, this is the kind of feedback I can use. We all want the same thing, to make it as good as we can! :)

To get back on your 3 points of attention:
1) You are completely right, didn't think of that. Will update the PR!
2 & 3) I see what you mean and I tried to do it without the default lang changed event. The problem is though that I need to inform the component to update all keys when the default language changes (like with a normal use updates all translations as well). I could "abuse" the event that is already there for the use to update all translations. But this could lead to some strange behaviour (e.g. waiting for a use to finish, but getting a default lang changed event in between).

pndewit:
Removed the default language to be 'en'. Moved the logics that I put into setDefaultLang (which is not really functionality for a setter function) to the getParsedResult function. Also updated the unit tests to reflect the changes.

ocombe:
Ok good, a few last changes and I'll merge it:
- fix the conflicts (merge master into your branch)
- update the new directive with the same thing that you did in pipe
- add documentation on the new event in the README.md

pndewit:
1. Conflicts are merged
2. Updated new directive and added some extra unit tests (also for the new subscription). This made me realise that moving the get default language translations from setDefaultLang to getParsedResult was wrong. Setting the default language should emit an event with the translations for the default language. In theory the setDefaultLang function should work about the same way as the use function. So I moved it back to setDefaultLang (but left out the part from getParsedResult where it set the default lang to 'en' of course).
3. Added documentation about the new event in the README.md

Thanks for your help!!

src/translate.directive.ts
```diff
@@ -27,6 +28,13 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
                 this.checkNodes(event.translations);
             });
         }
+
+        // subscribe to onDefaultLangChange event, in case the default language changes
+        if(!this.onDefaultLangChangeSub) {
+            this.onDefaultLangChangeSub = this.translateService.onDefaultLangChange.subscribe((event: DefaultLangChangeEvent) => {
+                this.checkNodes(event.translations);
```
ocombe: Changing the default lang doesn't mean that we're changing the current lang, so it shouldn't trigger a checkNodes

src/translate.pipe.ts
```diff
@@ -94,6 +95,16 @@ export class TranslatePipe implements PipeTransform, OnDestroy {
             });
         }
 
+        // subscribe to onDefaultLangChange event, in case the default language changes
+        if(!this.onDefaultLangChange) {
+            this.onDefaultLangChange = this.translate.onDefaultLangChange.subscribe((event: DefaultLangChangeEvent) => {
+                if(this.lastKey) {
+                    this.lastKey = null; // we want to make sure it doesn't return the same value until it's been updated
+                    this.updateValue(query, interpolateParams, event.translations);
```
ocombe: Same thing here, the default lang doesn't mean that it's the current lang

src/translate.directive.ts
pndewit: You are right that changing the default lang is not changing the current lang, but it could potentially change any of the values right?
I mean, we could implement something which only checks the translations that are not translated yet (displaying the translation's key). That would make it more efficient, but would also require some more effort.

ocombe: ok then it should call checkNodes without the translations argument, so that it tries to get the values from the current lang and fallback if necessary

pndewit: Ah, you are so right!! Will update it immediately! Thanks!

pndewit: FAILED TESTS:
  TranslateDirective
    ✖ should update the DOM when the default lang changes
      Chrome 54.0.2840 (Mac OS X 10.12.1)
    Expected 'TEST' to equal 'This is a test'.
        at Object.<anonymous> (webpack:///tests/translate.directive.spec.ts?f2bc:115:0 <- config/spec-bundle.js:72492:73)

Seems like it is not updating the keys now any more.

pndewit: Seems like I forgot to add tests for the pipe as well when the default lang changes. Added those to check if it does work there without the translations appended to the updateValue call and it does work there. Maybe an error in the checkNodes function in the directive?

ocombe: Yeah there's a prob for the directive, the tests don't wait for the DOM change for some reason, it works but the test fails. It might have to do with the fact that I change the DOM manually (bad!), I should use the renderer maybe

ocombe: Yes, you can do that, I'll take a look at that problem when I have the time, because we probably need a better dom management to support nativescript and such

pndewit:
@ocombe Updated PR, removed the translations argument from the DefaultLangChange events in the pipe and the directive.

pndewit:
It's ready to be merged. Let me know if I am missing something!

pndewit:
@ocombe Can this be merged?

pndewit:
@ocombe Is this on hold/waiting for something?

ocombe:
Sorry, been a bit busy, I'll try to do it this week

pndewit:
No worries, not trying to insult or push you. I was just wondering what the status was.

ocombe:
No problem :)

ocombe:
Ok I fixed the bug in the directive, the failing test will pass now.
I also fixed a few other things at the same time: it sets the default lang to the first lang that receives translations (if the default lang is undefined at that time), and the onTranslationChange event now emits for all setTranslation calls (not just for merges, I don't know why this was just for merge).
Also the directive now listens to onTranslationChange events.

ocombe:
All of this is available in the new 5.0.0 release.

pndewit:
Nice, great work!

PR #344, opened 2016-12-05
Directive

ocombe:
Fixes #31


src/translate.directive.ts
```diff
@@ -0,0 +1,90 @@
+import {Directive, ElementRef, AfterViewChecked, Input, OnDestroy} from "@angular/core";
+import {Subscription} from "rxjs";
+import {isDefined} from "./util";
+import {TranslateService, LangChangeEvent} from "./translate.service";
+
+@Directive({
+    selector: '[translate]'
```
ocombe: sure I'm gonna add ng-translate for now, but all of this will be renamed to ng-i18n once the name of the lib changes (soon !)

ocombe: Actually I'll add `ng2-translate`

PR #345, opened 2016-12-05
feat(Parser): rename DefaultParser to DefaultTranslateParser.

aitboudad:


ocombe:
Good point, thanks :)

PR #348, opened 2016-12-05
Make TranslationParser public

biesbjerg:


alexluecke:
I think member should be protected. Is there a reason you need to call parser outside of the base or derived classes?

biesbjerg:
Yes. Related ticket https://github.com/ocombe/ng2-translate/issues/347

Issue #325, opened 2016-11-18
ng2-translate with webpack does not work without copy-webpack-plugin


yogesh51:
**ng2-translate with webpack does not work without copy-webpack-plugin**
The following example https://github.com/ocombe/ng2-translate/tree/master/examples/webpack is not working without copy-webpack-plugin. It needs to be used webpack.config.js to copy the JSON files to the build folder. @ocombe  please update these (path as per the folder structure)
`var CopyWebpackPlugin = require('copy-webpack-plugin');
plugins: [
new CopyWebpackPlugin([ { from: 'src/i18n', to: 'dist/assets/i18n' } ])
]`

Environment:

* **ng2-translate version:** 4.0.0

* **Angular version:** 2.1.0-rc.5
 
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ] 
 
* **Language:** [all | TypeScript X.X | ES5]

ocombe:
Hello,

Would you like to do a PR for that ?

stephengeorgewest:
Created a pull request. #356

ocombe:
This has been merged, thanks.

PR #356, opened 2016-12-13
copy-webpack-plugin

stephengeorgewest:
pull request for "ng2-translate with webpack does not work without copy-webpack-plugin #325". I needed to figure this out in my own project, so here it is.


ocombe:
Thanks a lot ! :)

PR #379, opened 2016-12-30
add forChild()

atiertant:
TranslateModule can be imported multiple times: once per lazily-loaded bundle.
That is why there are two ways to create the module: TranslateModule.forRoot and TranslateModule.forChild.

forRoot creates a module that contains all the directives, the given loader, and the TranslateService service itself.

forChild creates a module that does not include the TranslateService service but a new given loader for the module (useful for splitting translation files).

When registered at the root, the module should be used as follows

```
@NgModule({
  imports: [TranslateModule.forRoot(Loader)]
})
class MyNgModule {}
```

For submodules and lazy loaded submodules the module should be used as follows:

```
@NgModule({
  imports: [TranslateModule.forChild(Loader)]
})
class MyNgModule {}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,21 @@ If you use a [`SharedModule`](https://angular.io/docs/ts/latest/guide/ngmodule.h
 export class SharedModule { }
 ```
 
-> Note: Never call a `forRoot` static method in the `SharedModule`. You will end up with multiple different instances of a service in your injector tree.
+> Note: Never call a `forRoot` static method in the `SharedModule`, use `forChild` instead. You will end up with multiple different instances of a service in your injector tree.
+
+```ts
+@NgModule({
+    exports: [
+        CommonModule,
+        TranslateModule.forChild({
+            provide: TranslateLoader,
+            useFactory: (http: Http) => new TranslateStaticLoader(http, '/assets/i18n/SharedModule', '.json'),
+            deps: [Http]
+        })
+    ]
+})
+export class SharedModule { }
+```
 
 ##### Configuration
 
@@ -250,7 +264,7 @@ To render them, simply use the `innerHTML` attribute with the pipe on any elemen
 	  // do something
 	});
     ```
-    
+
 #### Methods:
 
 - `setDefaultLang(lang: string)`: Sets the default language to use as a fallback
--- a/index.ts
+++ b/index.ts
@@ -2,7 +2,7 @@ import {NgModule, ModuleWithProviders} from "@angular/core";
 import {Http, HttpModule} from "@angular/http";
 import {TranslatePipe} from "./src/translate.pipe";
 import {TranslateParser, DefaultTranslateParser} from "./src/translate.parser";
-import {TranslateService, TranslateLoader, TranslateStaticLoader} from "./src/translate.service";
+import {TranslateService, TranslateLoader, TranslateStaticLoader, ModuleLoader} from "./src/translate.service";
 import {TranslateDirective} from "./src/translate.directive";
 
 export * from "./src/translate.pipe";
@@ -32,11 +32,30 @@ export class TranslateModule {
         useFactory: translateLoaderFactory,
         deps: [Http]
     }): ModuleWithProviders {
+        // var id = 'root';
         return {
             ngModule: TranslateModule,
             providers: [
                 providedLoader,
                 TranslateService,
+                ModuleLoader,
+                { provide: TranslateParser, useClass: DefaultTranslateParser }
+            ]
+        };
+    }
+    static forChild(providedLoader: any = {
+        provide: TranslateLoader,
+        useFactory: translateLoaderFactory,
+        deps: [Http]
+    }): ModuleWithProviders {
+        console.log('for child');
+
+        // var id = 'child';
+        return {
+            ngModule: TranslateModule,
+            providers: [
+                providedLoader,
+                ModuleLoader,
                 { provide: TranslateParser, useClass: DefaultTranslateParser }
             ]
         };
--- a/index.ts
+++ b/index.ts
@@ -38,7 +38,7 @@ export class TranslateModule {
             providers: [
                 providedLoader,
                 TranslateService,
-                ModuleLoader,
+                { provide: ModuleLoader, useValue: {uid: 'root'} },
                 { provide: TranslateParser, useClass: DefaultTranslateParser }
             ]
         };
--- a/index.ts
+++ b/index.ts
@@ -1,4 +1,4 @@
-import {NgModule, ModuleWithProviders} from "@angular/core";
+import {NgModule, ModuleWithProviders, Injector} from "@angular/core";
 import {Http, HttpModule} from "@angular/http";
 import {TranslatePipe} from "./src/translate.pipe";
 import {TranslateParser, DefaultTranslateParser} from "./src/translate.parser";
@@ -27,6 +27,12 @@ export function translateLoaderFactory(http: Http) {
     ]
 })
 export class TranslateModule {
+
+    static INJECTOR: Injector;
+
+    constructor(Injector: Injector) {
+      TranslateModule.INJECTOR = Injector;
+    }
     static forRoot(providedLoader: any = {
         provide: TranslateLoader,
         useFactory: translateLoaderFactory,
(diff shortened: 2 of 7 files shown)


ocombe:
Hi, could you explain what is the idea behind this PR ? Why is it needed, what does it offer ?

atiertant:
@ocombe oups, sorry I updated my comment

ocombe:
oh ok, so this allows you to have a loader for each of your modules, it might be a different type of loader if you want, and each loader has its own translations that it doesn't share with the other modules ?
I have to think about this, because it's a major change.
Allowing each module to load its own translations is one thing, but making all of them independent is another, I can already imagine people who will want those translations to be shared among module.

atiertant:
@ocombe that's it, this is the same idea than angular router, this allow lazy-loading...
how do you think moduled should share translations? only one big translation object merging new other from child modules (what about performance on large app) ?

ocombe:
I'm not saying it's a bad idea, it's just that I need to think about it before I merge this, because once it's in there is no going back ;-)

atiertant:
this could be improved a bit by adding a parent to each module, if there is no translation on it, try its upper parents...

ocombe:
Ok so I finally took the time to review this PR (sorry it took me so long).

At first I didn't get why you generated uids and why you moved a lot of things into the module loader. The reason is that we want "one" instance of the service, so that we keep one list of languages and translations, but we want different loaders. The loader being injected into the service, it couldn't work for sub-modules.
So you created a new service that gets registered into the main service, and when the pipe / directive need a translation, it gives the id to the main service so that it uses the local instance that uses the local loader.

But I don't like that because it complicates a lot the model, and it changes the signature of all the methods to add a new "uid" parameter.

I searched how we could achieve this modularity without changing the methods, here is what I think would be a better way to do it:
- forRoot defines a service / loader / directive / pipe and **store**
- this store is new, it will be used to share common stuff between each instance of the service
- forChild defines a service / loader / directive / pipe, but **no** store, the store is unique to the root module, it will be injected in each instance of the service
- since the service is local, it will use the loader loader, which means that the methods don't change in the pipe / directive and components that use the service
- the service will call this shared store when it wants global stuff (the list of languages, translations, ...)
- if you want totally independent modules that don't share the translations/languages/..., you can use forRoot in each independent module, it will create a local instance of the store

What do you think? It would be much simpler, don't you think?

SamVerschueren:
I totally agree with @ocombe. I think that's the way to go.

One remark though is when you have the same keys and different module translation files. For example, if we have this translation file for the root

```json
{
    "foo": "bar"
}
```

And for a feature module we have this translation file

```json
{
    "foo": "hello world"
}
```

The feature module will overwrite the value that we already have in the store and the `foo` translation key will map to `hello world` in every other module.

So what if we did it like Angular's injector. Create a store per feature module. If someone asks for a translation, it will look for it at the module store where it lives in. If the translation key could not be found it will navigate one up and asks the parent store for the translation and so on. It's just an idea though and not sure if it can be easily implemented. Just wanted to share the idea :).

atiertant:
@ocombe why not using multiple instance of service if it's simpler... does service would generate their own id to buid a tree in store or would you like all keys to be merged in a flat object? @SamVerschueren this is exactly what i was thinking...

atiertant:
@ocombe maybe store could have an option `mergeChild` when active all translations are merged in the same object else every service translations are stored in a tree (if a key does not exist on child, search recursively in parent)

ocombe:
Ok so I tried to implement my method, and wrote a test for it but... it fails.
I had 2 different modules, each one importing TranslateModule with forChild, and tried to define "useStore" to false, but both modules shared the same TranslateService instance, even if forChild provided it twice.
I asked Rob Wormald why that was the case, and apparently even if you use different modules, the providers get flatten at bootstrap and the last "provide" wins.
Obviously that's not the case for lazy loaded modules because they are not defined at bootstrap. They get their own injectors, which means that if you provide a service there, it will be a new instance. But since everything runs inside the same angular application, the new injector inherits from the root injector, and you can still get parent instances (if you don't provide a new one in the lazy loaded module).

So the only way to have different loaders / services in different modules will be to lazy load them (or use the `providers` property in your components, which creates a new injector too).

It means that you can do all of that, it will just not be possible to have 2 different instances in 2 different modules that are not lazy loaded.

atiertant:
@ocombe 
why would you like to use forChild without lazy loaded module?
do you mean it should call every loaders at the same time when asking for new language?

could you publish your code?
maybe we could use store instance in .forChild method to register the new loader.

my problem with lazy loaded module is to know who is the parent service instance.

ocombe:
You don't really have to use forChild if you don't use lazy loaded modules. You're right about the observables (on lang change), it's gonna be messy if the service is a new instance every time...

atiertant:
@ocombe so what's going on?

ocombe:
The store & forChild are available in the [ngx-translate branch](https://github.com/ocombe/ng2-translate/tree/ngx-translate) (along with a bunch of other things).
There is an "isolate" property that you can set to "true" if you want to isolate your service (it only works with lazy loaded modules, or if you use the "providers" property in a component).

I will publish a beta of the new ngx-translate branch soon.

ocombe:
New 6.0.0-beta.1 with forChild: https://github.com/ngx-translate/core/releases/tag/6.0.0-beta.1
Be careful it is based on the new ngx-translate branch, and a lot of things have changed.

atiertant:
@ocombe after some test with the 6.0.0-beta.1 i found some problems with this implementation:
* as there is one instance of translate service by module, we need to reconfigure it in all modules (with use, langs,...), we could move this values in store to fix this.
* with `isole: false` once a lang has been loaded by an other module, the current module can't load his own translations
* with `isole: true`, as there is one instance of store by module, we can't fix first problem and this is not possible to use parent module translations when no translation is defined for the key.(see below)

i give an sample of what we expect to do:

i have two module first have 2 translation key in 'en' lang:
```
{
  'title': 'welcome on my site',
  'section': 'home'
}
```

and the second lazy loaded by first one has only one key:
```
{
  'section': 'contact us'
}
```

each of them show on top page `title` and `section` so first on show:
```
welcome on my site
home
```
and second show:
```
welcome on my site
contact us
```
second module has no translation for `title`,so higher module's one is used

i think we should isolate module translations in same store instance

SamVerschueren:
Creating a sample demo with Angular CLI now and I noticed the same thing @atiertant already described. In a feature module where you configure `TranslateModule.forChild` with a custom loader, you again have to call `.use()` and configure the service. I think all the different service should actually share the same `TranslateConfig` or something. A quick example of that could be something like this

```js
@Injectable()
class TranslateConfig {

    lang$ = new ReplaySubject<string>();

    use(lang: string) {
        this.lang$.next(lang);
    }
}
```

That `TranslateConfig` is only created in the `forRoot` call so that it is shared between all the `TranslateService` instances. They can then subscribe to changes on the config object like the `lang$` observable. If a service changes the language, all the other service will also respond and load the correct language.

Just a quick idea :)

cocaux:
Hi, 
I need this functionality for my devs, do you know when it will be available in npm? 
Can I directly point my package.json to this repo? 

Thanks for your help

ocombe:
It's available in 6.0.0

cocaux:
Thanks for the quick answer :). I found some documentation in the README, but do you have a working example with a CustomLoader doing an http get?

ocombe:
Not really but you can take a look at the http loader source code: https://github.com/ngx-translate/http-loader/blob/master/src/http-loader.ts

cocaux:
I still have issue... I got a error caused by: No provider for TranslateStore!
My current set up is:
- in my module:
`TranslateModule.forChild({
      loader: {
        provide: TranslateLoader,
        useClass: CustomLoader,
        deps: [Http]
      }, isolate: true
    }),`
- my loader:
`export class CustomLoader implements TranslateLoader {
  constructor(private http: Http) { }
  getTranslation(lang: string): Observable<any> {
    return this.http.get('...?language='+lang)
      .map((res: Response) => res.json());
  }
}`
- in my main component in my module: 
`  constructor(translate: TranslateService) {
        translate.setDefaultLang('en');
        translate.use('en');
    }
`
I am missing something ? Should I do a `TranslateModule.forRoot()` at the root level ?
Many thanks for your help

cocaux:
@ocombe, I tried different set up, but I never saw the TranslateService using the CustomLoader set in the Translate.forChild. Is the TranslateService managed the different loaders? for me it looks like it always use the one set in the forRoot.

Many Thanks
Coco

mebibou:
@ocombe is `forChild` still supposed to work in `8.0.0`? I tried to use it with a different loader than the one in `forRoot`, and using `isolate: true` and `isolate: false`, but it doesn't load the translations at all. I think the idea of having children load their own translations but listening to `lang` change on the parent is crucial when using modules

PR #394, opened 2017-01-15
feat(TranslateParser): inject interpolateParams with currentLang.

aitboudad:
I need this for a custom plural parser
example: https://github.com/aitboudad/ng2-translate/blob/dev/src/translate.parser.ts#L58-L79

### DIFF
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -341,11 +341,11 @@ export class TranslateService {
         }
 
         if(translations) {
-            res = this.parser.interpolate(this.parser.getValue(translations, key), interpolateParams);
+            res = this.parser.interpolate(this.parser.getValue(translations, key, interpolateParams, this.currentLang), interpolateParams, this.currentLang);
         }
 
         if(typeof res === "undefined" && this.defaultLang && this.defaultLang !== this.currentLang) {
-            res = this.parser.interpolate(this.parser.getValue(this.translations[this.defaultLang], key), interpolateParams);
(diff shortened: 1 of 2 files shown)


src/translate.service.ts
```diff
@@ -286,9 +286,12 @@ export class TranslateService {
      * @param interpolateParams
      * @returns {any}
      */
-    public getParsedResult(translations: any, key: any, interpolateParams?: Object): any {
+    public getParsedResult(translations: any, key: any, interpolateParams?: any): any {
         let res: string|Observable<string>;
 
+        interpolateParams = interpolateParams || {};
+        interpolateParams.currentLang = interpolateParams.currentLang || this.currentLang;
```
ocombe: This is a problem, it will define `interpolateParams` even when there are no params. Why do you need the lang in the custom plural parser ?

aitboudad: now it should be ok, see the use case https://github.com/aitboudad/ng2-translate/blob/dev/src/translate.parser.ts#L58-L79

aitboudad:
ping @ocombe

SamVerschueren:
Why do you need the lang in the custom plural parser ?

aitboudad:
@SamVerschueren I'm using [messageformat](https://github.com/messageformat/messageformat.js), just looks at the provided [example](https://github.com/aitboudad/ng2-translate/blob/dev/src/translate.parser.ts#L58-L79) and you'll see the use case

SamVerschueren:
Okay, I see what you're doing. Seems like a valid use case.

PR #396, opened 2017-01-18
Adding useful LocalizableTranslateStaticLoader

ignaciolarranaga:
In some situations you need to handle different localizations for the same language.
Initially I thought it was cover by:

```
_translateService.setDefaultLang("en");
_translateService.use("en-UK");
```

but certainly not, it just ignore all the keys of en, if they are not present on en-UK.
So after reading some issues without a clear solution I came up with this idea and thought I might worth to be shared because it sounds to be a common scenario and a good approach.

The provided loader if used will look for both files and override the keys of the generic translation with the localizations, so for example if you have:
es.json: { "currency": "Moneda", "currency-symbol": "$" }
and 
es-UY.json: { "currency-symbol": "$U" } -> only currency symbol is going to be override
es-AR.json: { "currency-symbol": "ARS" }
es-UK.json: { "currency-symbol": "£" }

all the translations of es.json are going to be available but just the repeated ones overriden, which is the common scenario on application localization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,18 @@ export function createTranslateLoader(http: Http) {
 export class AppModule { }
 ```
 
+##### Localizations
+
+If you need to handle different localizations for the same language you can use the LocalizableTranslateStaticLoader. It allows you to manage different files (es.json, es-UY.json, es-US.json, ...) were each specific localization file overrides the keys of the default one.
+
+```ts
+export function createTranslateLoader(http: Http) {
+    return new LocalizableTranslateStaticLoader(http, './assets/i18n', '.json');
+}
+```
+
+Note: Remember to use the specific locale, e.g. `translate.use('es-UY');`
+
 #### 2. Init the `TranslateService` for your application:
 
 ```ts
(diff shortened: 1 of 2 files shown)


ignaciolarranaga:
@ocombe sure, give me a few hours to publish the package on npm and will get back to you.
Do you want me to include the static loader as it is now on the library for the future ?

ocombe:
No, you don't have to, if you want to take a look at what it will be, it's here: https://github.com/ngx-translate/http-loader/blob/master/src/http-loader.ts (I renamed it http loader).

But I will make a few changes again, because I want to add a new functionality: the format. Right now it only support json, but I want to make it possible to support other formats as well, so I'll probably add a new dependency to a "format" library as well.

ignaciolarranaga:
And don't you prefer a pull request with the class to http-loader ?, might have sense ? (I mean instead of a separate plugin on npm)

ocombe:
This is a specific behavior that I prefer to keep separate

ignaciolarranaga:
sure, will get back to you as soon as I get npm plugin ready.

ignaciolarranaga:
@ocombe I did publish it on npm, here is the link: https://www.npmjs.com/package/ng2-translate-localize

Thanks for the advice, comments and suggestions are welcome.

PR #412, opened 2017-02-03
fix: do not return object as translation

deeg:
Previous to this fix, if passing in a partial key as a translation, it would return the child object instead of handling it as a missing translation and returning the key.

**Previously:**

Translation file: `{"admin": {"list": "Admin List"}}`
`{{'admin' | translate}}` displayed [Object object] because it was getting `{"list": "Admin List"}`

**After Fix:**
Translation file: `{"admin": {"list": "Admin List"}}`
`{{'admin' | translate}}` will display 'admin' or go through the custom missing translation handler.

### DIFF
--- a/tests/translate.service.spec.ts
+++ b/tests/translate.service.spec.ts
@@ -117,6 +117,17 @@ describe('TranslateService', () => {
         mockBackendResponse(connection, '{}');
     });
 
+    it('should return they key when getting an object as the translation if supplying half of a key', () => {
+        translate.use('en');
+
+        translate.get('TEST').subscribe((res: string) => {
+            expect(res).not.toEqual({"SUBTEST": "Dit is een test"});
+            expect(res).toEqual('TEST');
+        });
+
+        mockBackendResponse(connection, '{"TEST": {"SUBTEST": "Dit is een test"}}');
+    });
+
     it("should return the key when you haven't defined any translation", () => {
         translate.get('TEST').subscribe((res: string) => {
             expect(res).toEqual('TEST');
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -333,9 +333,10 @@ export class TranslateService {
                 params.interpolateParams = interpolateParams;
             }
             res = this.missingTranslationHandler.handle(params);
+            return typeof res !== "undefined" ? res : key;
         }
 
-        return typeof res !== "undefined" ? res : key;
+        return typeof res === "string" ? res : key;
     }
 
     /**


tests/translate.service.spec.ts
```diff
@@ -117,6 +117,17 @@ describe('TranslateService', () => {
         mockBackendResponse(connection, '{}');
     });
 
+    it('should return they key when getting an object as the translation if supplying half of a key', () => {
+        translate.use('en');
+
+        translate.get('TEST').subscribe((res: string) => {
+            expect(res).not.toEqual({"SUBTEST": "Dit is een test"});
```
deeg: I know both of these checks are not needed, but I did it to illustrate what it would have been before fixing the bug. If you want me to remove the not.toEqual check let me know, and I will!

SamVerschueren:
Thanks for your effort. But imo this is desired behaviour. If we render `[object Object]`, you can easily detect that you are doing something wrong. For instance, what if we have the following translation file

```json
{
    "Admin": "Admin"
}
```

And in the view we have

```html
<h2>{{ 'Admin' | translate }}</h2>
```

So far so good! But if we refactor our translation file to something like this

```json
{
    "Admin": {
        "Admin": "Admin",
        "Role": "Role"
    }
}
```

The current version will display `[object Object]` because we forgot to refactor our view. Your version on the other hand will display `Admin`.

So unless @ocombe thinks differently, I only see downsides to this approach.

ocombe:
Yeah I'm not sure about this either. You're supposed to be the master of your translations, you should know that some key contains an object (and use `key.childKey`).
Also some people might want to get an object (with `get` or `instant`) in their service and do stuff with it, it could be the intended behavior...
Like @SamVerschueren said, your change limits those possibilities and introduces a breaking change. An object is not really a missing translation. I understand that what you really want is to be able to go through the missing translations handler (if someone gives you bad translations), but fixing those errors at runtime are probably not the way to go, you should fix them in dev before you ship your code.

SamVerschueren:
Oh, I totally forgot about the use case that you sometimes want to get an object back in your JavaScript with multiple translations so you can work with that one instead of getting values one by one.

Because of all these downsides, I'm going to close this PR. Feel free to provide other arguments why we should reconsider.

deeg:
@ocombe @SamVerschueren,

No problem if you don't want to take the change, I just want to explain a bit more of the scenario I was in.

I had a generic facet component where sometimes the value for the facet was a known translation key, and sometimes it was user input which I wanted to display directly as is. 

Without taking this change, you can never really have a spot which sometimes takes a translation key, and sometimes takes user input to be displayed directly, because there is no way to guarantee it won't clash with the beginning of a translation key.

If you guys are fine with that, I'm happy to work around my situation as I already have, by defining if the facet values should be translated or not.

Just another note, a user would still be able to get the raw translation object with `translationService. getTranslation`. I'm still not sure why you would ever want instant or the translation pipe to return anything but a string.

Thanks for taking the time to review and talk it out with me.

Issue #416, opened 2017-02-06
Build script failed in Windows


gilhanan:
* **I'm submitting a ...**
[X] bug report

`npm run build` running on Windows throw:

` 'cp' is not recognized as an internal or external command, operable program or batch file. `

PR #417, opened 2017-02-06
fix(build): cross platform cp

gilhanan:
Fix https://github.com/ngx-translate/core/issues/416

* Install npm package 'cash-cp' as devDependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "@types/jasmine": "2.5.41",
     "@types/node": "7.0.4",
     "awesome-typescript-loader": "3.0.0-beta.18",
+    "cash-cp": "0.2.0",
     "codelyzer": "2.0.0-beta.4",
     "commitizen": "2.9.5",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "@types/jasmine": "2.5.41",
     "@types/node": "7.0.4",
     "awesome-typescript-loader": "3.0.0-beta.18",
+    "cash-cp": "^0.2.0",
```
deeg: This should probably be hardcoded to a specific version like the other dev dependencies.

gilhanan: Thanks, pinned to specific version

SamVerschueren:
Just wondering what the cause of this issue is. Which plugin/dependency uses `cp` internally?

ocombe:
This PR was for v5, but I changed the package.json and it no longer uses cp in the npm commands, so I'm closing the PR since it's no longer an issue :)

PR #423, opened 2017-02-08
Dynamic inline variables for the translate directive

Verurteilt:
When you have dynamic params for translating  it seems to be the Directive doesn't detect the changes on the params, thus the text on the HTML element remains the same, example:

Dictionary :

`{ "HELLO": "hello {{value}}" }`

Component:

```
@Component({
  selector: 'app-example',
  templateUrl: `<p translate [translateParams]="{value: person.name}">HELLO</p>`,
  styleUrls: ['./example.component.scss']  
})

export class ExampleComponent implements OnInit {

	person: any;


	constructor(){}

	ngOnInit(){
		this.person = {
			'name': "Angular 2 translate"
		}

		this.changePerson();
	}

	changePerson(){
		this.changePerson = {
			'name': "This value will never appear on the screen"
		}
	}
}
```

I implemented the OnChanges Interface so it gets fired whenever you change the translateParams.

### DIFF
--- a/src/translate.directive.ts
+++ b/src/translate.directive.ts
@@ -56,11 +56,20 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy, OnChange
     }
 
     ngOnChanges(changes: any) { //Everytime an Input property is changed, Angular fires ngOnchanges
+<<<<<<< HEAD
         if(changes.translateParams) {//We only care if the translateParams changed
             this.lastParams = changes.translateParams.previousValue; //Self explanatory
             // this.translateParams = changes.translateParams.currentValue; //Self explanatory
             this.currentParams = changes.translateParams.currentValue; //Self explanatory
             this.checkNodes(); //We check the nodes to update to the new value
+=======
+        for (let propName in changes) {
+            if(propName === "translateParams") {//We only care if the translateParams changed
+                this.lastParams = changes.translateParams.previousValue; //Self explanatory
+                this.translateParams = changes.translateParams.currentValue; //Self explanatory
+                this.checkNodes(); //We check the nodes to update to the new value                
+            }
+>>>>>>> ca43e2c... Dynamic inline variables for the translate directive
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/translate.directive.ts
```diff
@@ -53,6 +53,16 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
         this.checkNodes();
     }
 
+    ngOnChanges(changes: any) { //Everytime an Input property is changed, Angular fires ngOnchanges
+        for (let propName in changes) {
+            if(propName === "translateParams") {//We only care if the translateParams changed
```
deeg: If we only care if the translateParams changed, why do a loop through all of the properties in changes? I would instead remove the for loop and just have:

```
if (changes.translateParams) {
    'contents of original if block'
}
```

Verurteilt: Habit, I will change it ASAP

Verurteilt:
Any update on this pull request? Would be cool if you check it out

deeg:
People are busy with work and life, I'm sure it will be looked at soon.

Can you please squash your commits to make it easier to accept when people have looked at it?

ocombe:
It's ok I can squash them when I merge :)
I'll try to review all the pending PRs this week end, I'll have some time

ocombe:
So I reviewed the PR, and wrote a test for it, and it turns out that it didn't work well with components using OnPush.
I used the @Input setter instead of ngOnChange (which I don't like to use in a lib since it's triggered by all changes), and referenced the changeDetectorRef like I did in the pipe (which probably fixed a few other bugs at the same time).
Anyway, thanks a lot for your PR, even if it looks like I didn't merge it, it's just that I'm not that good at git commands (and apparently running the merge squash command line doesn't merge the thing on github), but I used your PR anyway to fix the bug.
It will be available in the next release tomorrow.

Verurteilt:
@ocombe  Awesome, thank you, looking forward to update the library

Verurteilt:
@ocombe  I just tested the same scenario I have with the new release (6.0.0) and it's not working, the initial value of the variable gets rendered, but after i change it it doesn't re render the text.

I think we should re-open this.

ocombe:
Can you create a new bug? This one is a PR

Verurteilt:
Done

PR #426, opened 2017-02-13
docs(README): fix travis & npm badges

juristr:
Fixes the travis and npm badges. The npm badge image link should be correct, however it didn't load at the moment because there seems to be issues with the shields.io service. At least that's what Cloudflare told me :smile:.


ocombe:
thanks!

PR #441, opened 2017-02-22
docs(readme): add missing slash

jsayol:


ocombe:
Thanks, it was fixed with #445! :)

PR #445, opened 2017-02-24
docs(README): Fix asset path in example

SamVerschueren:
Noticed that this changes since `ng2-translate`. We need a trailing `/` in order for that to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ If you want to configure a custom `TranslateLoader` while using [AoT compilation
 
 ```ts
 export function createTranslateLoader(http: Http) {
-    return new TranslateHttpLoader(http, './assets/i18n', '.json');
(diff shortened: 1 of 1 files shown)


SamVerschueren:
Also, I noticed that the default example already uses an export function, so not sure if the [AoT](https://github.com/ngx-translate/core#aot) section still makes sense.

ocombe:
Well at least people won't be able to say that they didn't know :D
Thanks for the docs fix

SamVerschueren:
Hahaha, true :D!

In Angular-CLI, you always need an export function. The default `ng serve` doesn't work with a lambda expression either so it's definitely a good thing to just use a consistent example. For people who are using SystemJS or a custom webpack config and think to be a smarty pants by using a lambda, there is still the `AoT` section for them as well.

So let's just keep it like this :).

PR #448, opened 2017-02-26
fix(TranslateService): only make one request per lang

### LINKED ISSUE #397: three request on init
```
[ x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] feature request
```

**Current behavior**
when app inits it sends three request for one json file
**Expected/desired behavior**
it should send one

screenshot from 2017-01-18 11-58-40

* **ng2-translate version:** 5.0.0

* **Angular version:** 2.3.1
 
* **Browser:** [all ] 
 
* **Language:** [ TypeScript ]

deeg:
This stops multiple requests being made for the same lang file. It will now return the same observable when making duplicate requests for the same language.
Closes #397

### DIFF
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -47,6 +47,7 @@ export class TranslateService {
     private _currentLang: string;
     private _langs: Array<string> = [];
     private _translations: any = {};
+    private _translationRequests: any  = {};
 
     /**
      * An EventEmitter to listen to translation change events
@@ -230,7 +231,8 @@ export class TranslateService {
 
         // if this language is unavailable, ask for it
         if(typeof this.translations[lang] === "undefined") {
-            pending = this.getTranslation(lang);
+            this._translationRequests[lang] = this._translationRequests[lang] || this.getTranslation(lang);
+            pending = this._translationRequests[lang];
         }
 
         return pending;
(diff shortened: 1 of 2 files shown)


tests/translate.service.spec.ts
```diff
@@ -296,4 +296,19 @@ describe('TranslateService', () => {
         expect(browserCultureLand).toBeDefined();
         expect(typeof browserCultureLand === 'string').toBeTruthy();
     });
+
+    it('should not make duplicate requests', fakeAsync(() => {
+        let getTranslationCalls = 0;
+        spyOn(translate.currentLoader, 'getTranslation');
+        translate.currentLoader.getTranslation['and'].callFake(() => {
```
deeg: This is throwing a TS compilation error: `TS7017: Index signature of object type implicitly has an 'any' type.`

Looking for help on how to get rid of this. `.and` does not work either.

src/translate.service.ts
```diff
@@ -230,7 +231,8 @@ export class TranslateService {
 
         // if this language is unavailable, ask for it
         if(typeof this.translations[lang] === "undefined") {
-            pending = this.getTranslation(lang);
+            this._translationRequests[lang] = this._translationRequests[lang] || this.getTranslation(lang);
```
deeg: I can use a nested if here to only set it to this.getTranslation() if it is undefined. Let me know if you prefer that.

fulls1z3:
Any updates, guys?

PR #466, opened 2017-03-12
chore(TranslateDirective): don't import all of RxJS

jsayol:
It made no difference in this case since Subscription is only being used as a type, and the import gets removed altogether from the generated source. But better safe than sorry.


ocombe:
Thanks!

PR #509, opened 2017-04-12
fix(parser): treat '' as missing translation

### LINKED ISSUE #377: Directive not working with Angular Universal: node.textContent is undefined
**I'm submitting a ...**
```
[X ] bug report
[ ] support request
[ ] feature request
```

**Current behavior**
The `TranslateDirective` in Angular Universal does not work due to the following error:

    TypeError: Cannot read property 'trim' of undefined
    at TranslateDirective.checkNodes (.../node_modules/ng2-translate/bundles/ng2-translate.umd.js:759:51)

e.g. `<h1 translate>HELLO.WORLD</h1>`

**Expected/desired behavior**
Once the loader is loaded, string within a node should be translated.

Loading the [TranslateUniversalLoader](https://gist.github.com/ocombe/593d21598d988bf6a8609ba5fc00b67e) is loaded, add to a template (included via `templateUrl`) a node with a translate directive and the text string within the node (.e.g `<h1 translate>HELLO.WORLD</h1>`.

Start `server.js` and watch for errors in the terminal.

**What is the expected behavior?**
The methods `checkNodes` and `updateValue` of `TranslateDirective` expect each `node` to have a property called `.textContent : string`, which apparently does not exist server side.

Server side we do have however access to a property `.data : string`, which seems to behave exactly as `.textContent`.

Be referencing all calls to `textContent` with a fallback to `.data` solves the issue, like so:

    // checkNodes
    var content = (node.textContent || node.data).trim();
    ...
    node.originalContent = node.textContent || node.data;
    
    // updateValue
    node.originalContent = node.textContent || node.data;
    if(node.textContent) node.textContent = _this.key ? node.currentValue : node.originalContent.replace(key, node.currentValue);
    else node.data = _this.key ? node.currentValue : node.originalContent.replace(key, node.currentValue);

Translate Directive can't be used server side as it is

* **ng2-translate version:** 5.0.0

* **Angular version:** 2.2.3
* **angular/platform-server**: 2.2.3,
* **angular2-platform-node**: 2.1.0-rc.1,
* **angular2-universal**: 2.1.0-rc.1,
* **angular2-universal-polyfills**: 2.1.0-rc.1
* **angular2-express-engine**: 2.1.0-rc.1
 
* **Node:** 7.3.0
 
* **Language:** TypeScript

### LINKED ISSUE #487: ngx-translate-extract translation placeholders should not suppress falling back to the default language
```
[x] feature request
```

**Current behavior**

We use ngx-translate-extract to keep translation files in sync, which automatically inserts a mapping to "" for new keys, which is convinient for the translator, but suppresses the ngx-translate fallback to the default language. 

**Expected/desired behavior**

To handle pending translations more gracefully, it would be nice if the translation placeholder inserted by ngx-translate-extract would cause a fallback to the default language at runtime.

I therefore propose that `""` be considered undefined by `TranslateDefaultParser.getValue()`. 

Default language:
```
{
  "My Tasks": "My Tasks"
}
```

Used Language:
```
{
  "My Tasks": ""
}
```

Expression `{{'My Tasks' | translate}}` should yield `"My Tasks"` rather than `""`.

Sometimes our translators take a little time, and it is better to serve our testers/users with a label in the wrong language than no label at all. 

* **ngx-translate version:** 6.0.0
* **Angular version:** 2.x.x
* **Browser:** all

bedag-moo:
because ngx-translate-extract defaults translations to '', and untranslated tests should not suppress falling back to the default language

Fixes #487

### DIFF
--- a/src/translate.parser.ts
+++ b/src/translate.parser.ts
@@ -41,7 +41,7 @@ export class TranslateDefaultParser extends TranslateParser {
         key = '';
         do {
             key += keys.shift();
-            if(isDefined(target) && isDefined(target[key]) && (typeof target[key] === 'object' || !keys.length)) {
+            if(target && target[key]  && (typeof target[key] === 'object' || !keys.length)) {
                 target = target[key];
                 key = '';
             } else if(!keys.length) {


SamVerschueren:
I don't agree on this one. The translation might be left blank on purpose.

I don't see the link with #377...

bedag-moo:
Sorry, I referenced the wrong issue. It's #487.

bedag-moo:
Why would a translation be left empty on purpose?

SamVerschueren:
One example where you might want to have a blank translation is the following (and it happens to be the case in an application of mine).

A user can track it's health data with the application. This data could be either the weight of the user and the number of steps he did that day. All of the data is dynamic, the only thing that you know is the value and the type. Now you can easily just loop over that list like this

```html
<ul>
    <li *ngFor="let datapoint of datapoints">
        {{ datapoint.value }} {{ 'datapoint.' + datapoint.type + '.unit' | translate}}
    </li>
</ul>
```

In case of the weight, the list item will be rendered like `76 kg` for instance. But steps don't have a unit so that translation will just be empty and my translation will look like this.

```json
{
    "datapoint": {
        "weight": {
            "unit": "kg"
        },
        "steps": {
            "unit": ""
        }
    }
}
```

If the value is set to an empty string, in my opinion this doesn't mean the translation doesn't exist and it should definitely not use a fallback.

bedag-moo:
... which would be bad because it would fall back to showing the key if the default language has an empty translation, too.

Ok, you've convinced me :-)

PR #513, opened 2017-04-17
Update app.component.ts

danielsogl:
Set Platform Language automatically


ocombe:
You could use this method "`getBrowserLang()`: string | undefined: Returns the current browser lang if available, or undefined otherwise"

MTechDE:
But when you run it on a Smartphone you should use this way.
I had sometimes problems with the getBrowserLang method.

Issue #444, opened 2017-02-23
Lazy loaded module is not adding the new custom translateLoader


jlberrocal:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**
When a lazy loaded module is being loaded it triggers the factory for create the new TranslateHttpLoader however it is not adding the translations in the file

**Expected/desired behavior**
Load the translations of the lazy loaded module

**Reproduction of the problem**
I can't provide a plnkr because but this is the code that i'm using
```
export function  translateFactory(http: Http) {
        return new TranslateHttpLoader(http, '/metro-apps/maintenances/i18n/', '.json');
}
[...]
TranslateModule.forChild({
    loader: {
        provide: TranslateLoader,
        useFactory: translateFactory,
        deps: [Http]
    }
})
```

**Please tell us about your environment:**
Kubuntu 16.04

* **ngx-translate version:** core=6.0.0 loader: 0.0.3

* **Angular version:** 2.4.7
 
* **Browser:** all

dhardtke:
Might be the same issue that I have: https://github.com/ngx-translate/core/issues/425?

jlberrocal:
Yes is similar, I try your solution but it didn't work for me

k-schneider:
Just bumped into this as well.

I wanted my app module to fetch translations from "/assets/i18n/<lang>.json" and my lazy loaded modules to fetch from "/assets/i18n/<feature>/<lang>.json" but was unable to get this to work.

The custom loader configured by forChild(...) was ignored.

jlberrocal:
what @k-schneider describes is exactly the same that i reported

atiertant:
the problem is that you must reconfigure the service in all lazy loaded modules but if you didn't set `isolate` to `true` and the lang has already been loaded by an other module, lang is not reloaded...

but if you set `isole` to `true` lazy loaded modules are totaly independant so if you would like use translations from parent modules and new others for this modules and its childs only, this is not working...

jlberrocal:
@atiertant i can't set isolate to true because i'll need the parent translations, i was expecting to be able to add more translations based on my current module (lazy)

jlberrocal:
will be nice to take advantage of lazy loading even with the translations, so we can add the needed translations for the module without lose the main (root) translations. actually i'm trying with this approach
```
[...]

Issue #425, opened 2017-02-11
.forChild() usage with lazy loading routes


dhardtke:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => check the FAQ and search github for a similar issue or PR before submitting
[x] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**
`AppModule`
```typescript
export function createTranslateLoader(http: Http) {
    return new TranslateHttpLoader(http, "/i18n/", "/index.json");
}
imports: [
    TranslateModule.forRoot({
            loader: {
                provide: TranslateLoader,
                useFactory: createTranslateLoader,
                deps: [Http]
            }
        }),
]
```
`SharedModule`
```typescript
imports: [
    //TranslateModule
],
exports: [
    //TranslateModule,
]
```

`AdminModule`
```
export function createTranslateLoader(http: Http) {
    return new TranslateHttpLoader(http, "/i18n/", "/admin.json");
}
imports: [
    TranslateModule.forChild({
            loader: {
                provide: TranslateLoader,
                useFactory: createTranslateLoader,
                deps: [Http]
            }
        }),
]
```

**Expected/desired behavior**
I would like to load additional JSON files per component / module, i.e. an admin.json for the administration panel of our app.
I would like to import one TranslateModule inside of `SharedModule` and configure it allow all modules that don't explicite call .forChild().

Both doesn't work, it simply doesn't load the admin.json and hence doesn't translate strings defined in that file.

**Please tell us about your environment:**

* **ngx-translate version:** 6.0.0-beta.1

* **Angular version:** 2.4.7
 
* **Browser:** [Firefox 51.0.1] 
 
* **Language:** [TypeScript 2.0]

neolanders:
Since I didn't found any plunker working with ngx-translate librairy and I also add some difficulties to manage to make it work with LoadChildren,
I've setup a way that work pretty well for me:

_I've created two SharedModules, (one for lazyLoading and one for the other part of my application)_

**SharedLazyModule for lazy loading content:**

```
@NgModule({
  imports: [
    HttpModule,
    CommonModule,
    TranslateModule.**forChild**({}),
  ],
  exports: [
    CommonModule,
    TranslateModule
  ]
})
export class SharedLazyModule {}
```


**SahredModule for App**

```
// AoT requires an exported function for factories
export function HttpLoaderFactory(http: Http) {
   return new TranslateHttpLoader(http, './assets/i18n/', '.json');
[...]

PR #531, opened 2017-05-02
Fix service configuration with isolate = true

atiertant:
no more need to repeat `setDefaultLang` and `use` on every service in lazy loaded modules

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,18 +30,18 @@
   "typings": "index.d.ts",
   "homepage": "https://github.com/ocombe/ngx-translate",
   "peerDependencies": {
-    "@angular/core": ">=2.0.0 || >=4.0.0-beta.0"
+    "@angular/core": ">=2.0.0"
   },
   "devDependencies": {
-    "@angular/animations": "4.0.0-rc.6",
-    "@angular/common": "4.0.0-rc.6",
-    "@angular/compiler": "4.0.0-rc.6",
-    "@angular/compiler-cli": "4.0.0-rc.6",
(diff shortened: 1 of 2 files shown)


mebibou:
I think this would solve #602 , #444 and #425 so seems like an important fix, could this be reviewed and merged?

ocombe:
Hmm so what does it change actually? Could I get some tests to make sure that it fixes the problem ?

mebibou:
@ocombe I don't know if it works, but the idea of not having to sync the children modules with the root one is clearly a good one. As said in [my comment on another PR](https://github.com/ngx-translate/core/pull/379#issuecomment-335134394), the current `forChild` doesn't work unless we call `use(lang)` in every children, then sync `lang` changes between `root` and children. It's very hard to maintain and doesn't work out of the box

nickwinger:
When will this be released ?

ocombe:
Once the PR has some tests :)

nickwinger:
so, no tests after 14 days, i guess i will stick with the workaround, or provide my own translation service...
thx

xTaz1:
Any updates here ?

darkurse:
This fix would be really welcome

rmccabe24:
Guys, are you going to merge this change into the main branch?  We could really use this useful change!

Krishnakanth94:
can any one please review and merge this pull request.Thanks in Advance.

Quinozio (approved):

JulianSanio:
Anybody who could resolve the conflict and approve this pr?

PR #532, opened 2017-05-02
Allow deepMerging of objects

jlberrocal:


ocombe:
I didn't want to depend on a library, so I copied the content of the deepMerge package inside the util file. Thanks for the PR, it will be available in the next version.

jlberrocal:
thanks @ocombe after my PR I thought you'd tell me that XD but what you did is the best approach

PR #535, opened 2017-05-03
forChild with isolate = true recursive translations

atiertant:
merge parent translations in current service instance


jlberrocal:
@atiertant just a quick question why are you using "self"? you are using fat arrow functions, so you'll never lose the context

atiertant:
@jlberrocal you're right...

ocombe:
Hello, thanks for the PR, however I don't think that should work like that with isolate = true, that's more a behavior that I would expect from a module that is **not** isolated.
For me, isolated means that it doesn't use anything from the parent...
I'll take the code from your PR, but change it so that it uses a new config parameter other than isolate.

ocombe:
Ok now I see why you want this for isolate only, because it has an independent loader, which is not the case otherwise.
I'm still adding a new option, because I don't think that it should always merge translations.

atiertant:
any ETA?

ocombe:
Closing this as there is now a merge option for translations, it should cover your use case

PR #538, opened 2017-05-16
Update README.md for @biesbjerg/ngx-translate-po-http-loader

AilisObrian:
@biesbjerg/ngx-translate-po-http-loader


ocombe:
good catch, thanks

PR #545, opened 2017-05-25
Fix and simplify deep merging

jlberrocal:
@ocombe AFAIK you haven't updated the NPM package yet, well... the deepMerge repo was not working as expected XD so i created a new function that is working (at least for me)


jlberrocal:
@ocombe any idea of when is this going to be merged? currently i'm needing to much this feature, and i don't want to have the tarball into my project

ocombe:
The difference is that it doesn't merge arrays ?

jlberrocal:
well it merge it for sure, but JS have some really weird behaviors with the objects taken from JSON files, that is fixed doing the JSON.parse(JSON.stringify())

PR #552, opened 2017-05-31
Support Angular 4.2.x OpaqueToken -> InjectionToken 

xmlking:
Replaced OpaqueToken with InjectionToken to make it compatible with Angular 4.2.x


ocombe:
thanks !

Issue #150, opened 2016-06-20
Feature: pluralization


gjuchault:
Hi,

Have you been thinking about implementing pluralization ?
https://angular-translate.github.io/docs/#/guide/14_pluralization

cschroeter:
We are using the [i18nPluralPipe](https://angular.io/docs/ts/latest/api/common/index/I18nPluralPipe-class.html) to solve this. 

**Html**
```html
<div>
  {{ messages.length | i18nPlural: messageMapping | translate :{count: messages.length} }}
</div>
```
**Typescript**
```javascript
class MyApp {
  messages: any[];
  messageMapping: any = {
    '=0': 'msg.none',
    '=1': 'msg.singular.',
    'other': 'msg.plural'
  }
  // ...
}
```
**Json**
```json
{
  "msg.none": "You have no new messages",
  "msg.singular": "You have one new message",
  "msg.plural": "You have {{count}} new messages"
}
```

gjuchault:
@cschroeter This might work for english, french etc. but russian will cause you some troubles

ocombe:
This is something that might be worth adding if the implementation of angular 2 doesn't work for me, but it's not a priority.
It would make a nice PR though.

Yimiprod:
Not working anymore since RC4 (https://github.com/angular/angular/issues/9826)

DethAriel:
While there is no built-in mechanism to do that yet, I came up with my own. If someone finds time to polish the thing, add unit tests, lang-change/translation change events and finally create a pull-request, that would be just awesome :)

```ts
import { ChangeDetectorRef, Injectable, Pipe, PipeTransform } from '@angular/core';
import { I18nPluralPipe, NgLocalization } from '@angular/common';

import { TranslateService } from 'ng2-translate';

/**
 * Maps a numeric value to a pluralized translated string.
 *
 * ## Usage
 *
 * expression | conPluralate:'PATH.TO.TRANSLATION.MAPPING'
 *
 * ## Example
 *
 * ```ts
 * @Component({
 *   selector: 'my-inbox',
 *   template: `
 *     <div>
 *       {{ messages.length | pluralate:'INBOX.MESSAGES_PLURAL_CATEGORIES' }}
 *     </div>
 *   `,
 * })
 * class MyInbox {
 *   messages: any[];
 *   // ...
 * }
 * ```
 *
 * While the translation file is:
 *
 * ```json
 * // en.json
 * {
 *  "INBOX": {
[...]

PR #553, opened 2017-06-01
Support pluralization using messageformat.js

lephyrus:
This PR aims to add support for the ICU message format syntax with the help of messageformat.js. I tried to find a way of integrating this with the current ngx-translate framework in a minimally invasive way. Here's the idea:

* Extend `TranslateParser.interpolate()` to not only support string interpolation but also function interpolation: `interpolate(fn, params) => fn(params)`
* Add a new configurable `TranslateCompiler` component which allows pre-processing translation values. This is an optional step and does nothing by default.
* Add a `TranslateMessageFormatCompiler` which compiles the translation values (expected to use the ICU syntax) into interpolation functions using messageformat.js.

Corresponding issue: https://github.com/ngx-translate/core/issues/150

**Edit:** If the dependency on messageformat.js is a concern, `TranslateMessageFormatCompiler` could easily be pulled in as a separate package of course, same as the "loaders". Only the first two points above affect the core. I guess this comes down to how popular one expects the ICU syntax to be with ngx-translate users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,15 +74,16 @@ export class SharedModule { }
 
 When you lazy load a module, you should use the `forChild` static method to import the `TranslateModule`.
 
-Since lazy loaded modules use a different injector from the rest of your application, you can configure them separately with a different loader/parser/missing translations handler.
+Since lazy loaded modules use a different injector from the rest of your application, you can configure them separately with a different loader/compiler/parser/missing translations handler.
 You can also isolate the service by using `isolate: true`. In which case the service is a completely isolated instance (for translations, current lang, events, ...).
-Otherwise, by default, it will share its data with other instances of the service (but you can still use a different loader/parser/handler even if you don't isolate the service).
+Otherwise, by default, it will share its data with other instances of the service (but you can still use a different loader/compiler/parser/handler even if you don't isolate the service).
 
 ```ts
 @NgModule({
     imports: [
         TranslateModule.forChild({
             loader: {provide: TranslateLoader, useClass: CustomLoader},
+            compiler: {provide: TranslateCompiler, useClass: CustomCompiler},
             parser: {provide: TranslateParser, useClass: CustomParser},
             missingTranslationHandler: {provide: MissingTranslationHandler, useClass: CustomHandler},
             isolate: true
@@ -342,6 +343,16 @@ Once you've defined your loader, you can provide it in your configuration by add
 })
 export class AppModule { }
 ```
+
+#### How to use a compiler to preprocess translation values
+
+By default, translation values are added "as-is". You can configure a `compiler` that implements `TranslateCompiler` to pre-process translation values when they are added (either manually or by a loader). A compiler has the following methods:
+
+- `compile(value: string, lang: string): string | Function`: Compiles a string to a function or another string.
+- `compileTranslations(translations: any, lang: string): any`:  Compiles a (possibly nested) object of translation values to a structurally identical object of compiled translation values. 
+
+Using a compiler opens the door for powerful pre-processing of translation values. As long as the compiler outputs a compatible interpolation string or an interpolation function, arbitrary input syntax can be supported.
+
 
 #### How to handle missing translations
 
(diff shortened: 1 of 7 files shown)


santialbo:
Can we get some feedback from the project matainers? This PR is addressing a major concern for many of the library users and no one has said anything in 2 weeks.

SamVerschueren:
I think it would be best to pull this in a separate package. People who don't need it shouldn't be dependent on messageformat. On mobile now so can't really dig in the code. I'll do my best to take a look in the coming days.

lephyrus:
@ocombe  @SamVerschueren Not to appear impatient, but I have implemented what I consider to be a careful solution to an issue that was tagged with "help wanted". It's been 3 weeks without feedback - is there still interest in this at all?

santialbo:
@lephyrus is there a way this could be pulled in a separate package somehow?

ocombe:
@lephyrus yes, this would be a too big change for this lib, it should probably be in a separate package. If there's something that's missing in the lib to allow you to do that as an external package, it could be done, but I don't want to pull an external lib as a dependency like that

lephyrus:
@ocombe Thanks for the feedback (and for this library). I've anticipated your concern, which is why I've already outlined in the description above that yes, the TranslateMessageFormatCompiler can be pulled in from a separate package. The first two points (function interpolation, compile step) affect the core, however. Since I've structured the commits accordingly, could you tell me if you'd be willing to merge the first two commits? I could then go about hosting the extra package.

ocombe:
yes, the compiler part can be merged, it can be useful to other people who want to write extensions as well

lephyrus:
@ocombe Ok, great. I do think the compiler in combination with allowing interpolation functions (not just interpolation strings) has a lot of potential.

I'll adapt the PR shortly.

lephyrus:
@ocombe I have brought the PR up-to-date as per our discussion. I don't have the package for `TranslateMessageFormatCompiler` ready but I'm hoping you can add a link once it is and this has been released.

Note that I find type guarantees to be somewhat murky around translations - firstly because translation objects are of the `any` type, and secondly because `getValue()` is used both to retrieve translation values as well as translation parameters. Since that's already the way it's documented in the README, I've changed the return type of `getValue()` to `any`. Improving type safety would need a more substantial refactoring, in my opinion.

Finally, I wasn't quite sure where in the README to put the notes about the compiler. Feel free to change the documentation as you see fit.

alexxxnf:
Any updates on this PR?

lephyrus:
@alexxxnf From my perspective, the code is ready. But I'm certainly not going to invest the time to set up a separate package for `TranslateMessageFormatCompiler` unless this is merged. I can't force the maintainers to be responsive, nor do they have to accept contributions. 🤷‍♂️

alexxxnf:
@lephyrus I got your point. We'll wait :)

lephyrus:
I guess what we can do is ping @ocombe - it's been a while, after all.

ocombe:
Yes sorry I was on holidays :)

lephyrus:
@ocombe Well, that was quick. Sorry to hear the holidays are over though... :wink: I'll be in touch once I have a link to the compiler package.

njulhb:
Hi @lephyrus , I find there is an issue that the strings are not compiled by the compiler when resource files are loaded slowly. I have found the root cause and submit a change in https://github.com/ngx-translate/core/pull/659. Could you please help me review the code change or fix the issue by yourself? 
I really need the fix in ngx-translate. 
Best Regards,
Haibo

imfarhadk:
@lephyrus, any timeline to publish the NPM package for TranslateMessageFormatCompiler.

thanks
farhad

lephyrus:
@imfarhadk I won't have time this weekend, but probably next. Let's say within two weeks from now. Knowing that some people want to use this helps, thanks.

lephyrus:
I've published the package:
https://www.npmjs.com/package/ngx-translate-messageformat-compiler

It's my first NPM package, so let me know if something's wrong. Also, publishing a few lines of code as an Angular library really is insanely more complicated than it should be...

PR #562, opened 2017-06-09
feat: Parser API support

mattschlitz:
This tweak to the API allows for non-string constructs to be passed as keys and have them be interpreted against the set of translations. Also, it simplifies the API in the translation service.

This will allow for a parser that can handle interpretable (and recursive) constructs like the following:

```
export class TranslatableToken {
    constructor(public key: string, public params?: any) {
    }
}

export class AlternativeParser extends TranslateDefaultParser {

    public parse(translations: any, key: any, params?: any): string {
        if(key instanceof TranslatableToken) {
            if(key.params) {
                if(params) {
                    params = Object.assign(this._parseTranslatableValues(translations, key.params), params);
                } else {
                    params = this._parseTranslatableValues(translations, key.params);
                }
            }
            key = key.key;
        }
        return super.parse(translations, key, params);
    }

    private _parseTranslatableValues(translations: any, values: any): {} {
        let ret: any = {};
        for(let key in values) {
            if(values.hasOwnProperty(key)) {
                let value: any = values[key];
                if(value instanceof TranslatableToken) {
                    value = super.parse(translations, value.key, this._parseTranslatableValues(translations, value.params));
                }
                ret[key] = value;
            }
        }
        return ret;
    }
}
```

### DIFF
--- a/src/translate.parser.ts
+++ b/src/translate.parser.ts
@@ -18,7 +18,7 @@ export class TranslateDefaultParser extends TranslateParser {
     templateMatcher: RegExp = /{{\s?([^{}\s]*)\s?}}/g;
 
     public parse(translations: any, key: any, params?: any): string {
-        return this.interpolate(TranslateDefaultParser.getValue(translations, key), params);
+        return this.interpolate(this.getValue(translations, key), params);
     }
 
     protected interpolate(expr: any, params?: any): string {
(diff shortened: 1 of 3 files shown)


mattschlitz:
After further examining the ramifications of this change, I think it would take a larger change to make this work so I'm closing the request.

PR #586, opened 2017-07-02
added type definition of the overloaded get result

beenotung:


beenotung:
close this pull request and target for v6.0.1

beenotung:
I'd messed up the usage of tag and branch, reopening this pull request

ocombe:
Hello, this won't work, the file "src/translate.service.d.ts" will be overwritten when we build.
Why do you feel like you need this change?

beenotung:
just for the sake of type completeness, otherwise, I'll need to cast the result.

If it's included in the latest build I'll try to update the version in package.json and try again

SamVerschueren:
Like @ocombe mentioned, this will be overwritten and isn't going to be pushed to `npm`. If you want to do what you're trying to do, you'll have to change [this line](https://github.com/ngx-translate/core/blob/master/src/translate.service.ts#L380) in the `TranslateService` to something like this.

```js
/**
 * Gets the translated value of a key
 * @param key 
 * @param interpolateParams
 * @returns {any} the translated key, or an object of translated keys
 */
public get(key: string, interpolateParams?: Object): Observable<string | any>;

/**
 * Gets the translated value of an array of keys
 * @param key 
 * @param interpolateParams
 * @returns {any} the translated key, or an object of translated keys
 */
public get(key: string[], interpolateParams?: Object): Observable<string | any>;

public get(key: string | string[], interpolateParams?: Object): Observable<string | any> {
    // Implementation goes here
}
```

This is how TypeScript implements method overloading. This way, when you type `get`, it will show 2 options instead of 1. The actual implementation method is hidden from autocompletes.

beenotung:
Didn't know this feature before, thanks for sharing.

However, your impl cannot restrict the output to be `Observable<string>` when the key is a `string`.
Is it the limitation of the typescript compiler or is it valid to write the annotation with different result type?

SamVerschueren:
You can write as many overloads as you want.

SamVerschueren:
@erhimanshugarg The example in the README is clear enough https://github.com/ngx-translate/core#4-use-the-service-the-pipe-or-the-directive.

```json
{
    "HOME": {
        "HELLO": "hello {{value}}"
    }
}
```

```js
translate.get('HELLO', {value: 'world'}).subscribe((res: string) => {
    console.log(res);
    //=> 'hello world'
});
```

PR #603, opened 2017-07-14
feat-TranslateService: default language translation configurable

mmwlada:
feat(TranslateService): default language translation when current language translation is missing is now configurable.

As it is now, if KEY isn't found in current language translation, translation from default language is returned. If there isn't KEY in default language, MissingHandler is invoked. 

This PR aims to make translation to default language configurable.

In some instances we don't want default translation to mix with current translation, and we want MissingHandler to jump in instead.


ocombe:
Hello, sounds good, could you just update the readme to add info about this new config option please?

PR #605, opened 2017-07-17
dont change the language if the language given is already selected

macjohnny:


macjohnny:
if you agree, i would add the same behavior to the `changeLang()` method

ocombe:
Hello, sound good thanks !

macjohnny:
@ocombe awesome, thanks for the superfast merge and release!

PR #610, opened 2017-07-20
Update translate.service.ts

prophetw:
if lang.json get the key but value is '' 
first will use default lang to get translations
if default lang do not have the key or value is ''
will trigger missingHandler


prophetw:
if translate find key with value '' 
should handle this not just return ''
angular-translate do like this  
if value is '' it will use fallback language translations as result

ocombe:
Hello and thanks for the PR, but I can see a few use cases where an empty translation can actually be a valid result, I'm closing this because it would be a big breaking change for those people with no workaround...

PR #633, opened 2017-08-13
Fix typo in the README

gise88:
The CommonModule and the TranslateModule needs to be imported, not exported


ocombe:
Hello, sorry but this example explains how to use a shared module to export those modules everywhere, it's meant to be "export".

PR #643, opened 2017-08-28
Fix Code Sample in README

SyntaxStacks:
These code samples seem to have been swapped. hope this helps.


ocombe:
Hello, thanks for your PR, but I'm not sure that this change is necessary

PR #650, opened 2017-09-04
Fix TranslateService.get for arrays; ensure correct key-value pairs

Almar:
Object.merge does not ensure order. When some of the translations are
still ‘missing’ but others are not, the array of the translations are
no longer in the same order of the original ‘key’ array. With the
original implementation the translations can get mixed up.


ocombe:
Closing this as this part of the code has been changed a lot. If you think that there's still an issue there, feel free to reopen the PR

PR #659, opened 2017-09-07
Fix when the pending state is true, the translation is not compiled by the compiler.

njulhb:
Hi, the change fixes an issue that when the translation is in pending state, the compiler will not compile the translation and the strings on the first loaded web page are not compiled.

### DIFF
--- a/lib/src/translate.service.ts
+++ b/lib/src/translate.service.ts
@@ -404,7 +404,7 @@ export class TranslateService {
                     observer.error(err);
                 };
                 this.loadingTranslations.subscribe((res: any) => {
-                    res = this.getParsedResult(res, key, interpolateParams);
+                    res = this.getParsedResult(this.compiler.compileTranslations(res, this.currentLang), key, interpolateParams);
                     if(typeof res.subscribe === "function") {
                         res.subscribe(onComplete, onError);
                     } else {


lephyrus:
Hey @njulhb, thanks for catching this and bringing it to my attention. Couple of points:
* I think the call to `compileTranslations()` should be in the `onComplete()` handler. Where you have it currently, the code assumes that this could still be another observable. So more like this:
```ts
                let onComplete = (res: string | Function) => {
                    res = this.compiler.compileTranslations(res, this.currentLang);
                    observer.next(res);
                    observer.complete();
                };
```
* You should use `npm run commit` to generate the commit message.
* I am not a maintainer, so I can't merge your request (or my own).

njulhb:
Hi @lephyrus , thank you for your comments. I have closed the PR. So could you please help fix the issue and submit your code change since you're the owner of feature "TranslateCompiler"?

lephyrus:
My feedback to you was from one user of ngx-translate to another. I'm not the "owner" of anything, I just contributed a little bit of code in my free time. You should do the same and pursue this PR. You've already done the hard work: figuring out what's wrong.

njulhb:
Hi @ocombe, could you please help review the fix and give some comments?

lephyrus:
I've had two additional reports of users running into this. Unfortunately it was an oversight when I implemented the compiler support into the core. I did not realize that the `TranslateService::get()` method does it's own parsing (and should do it's own compiling) if translation loading is pending at the time it's called.

Any objections to merging this, @ocombe?

bjorkblom:
Hi,

I'm facing the same issue that the translations doesn't compile when reloaded the page described in #710. The reason for this is is exactly what @njulhb has fixed in this PR, but since the `this.getParsedResult` method needs the compiled translations, the `this.compiler.compileTranslations` method in the `onComplete` callback runs to late.

So the fix @njulhb did in the first place actually fixes the issue, and then when this PR was updated to run `this.compiler.compileTranslations` in the `onComplete` callback it was broke again and all translations prints the actual functions the should be interpolated first.

src/translate.service.ts
```diff
@@ -391,6 +391,7 @@ export class TranslateService {
         if(this.pending) {
             return Observable.create((observer: Observer<string>) => {
                 let onComplete = (res: string) => {
+                    res = this.compiler.compileTranslations(res, this.currentLang);
```
bjorkblom: Remove line

bjorkblom:
Can this fix be merged?

njulhb:
Hi @bjorkblom , please merge the fix

bjorkblom:
Hi @njulhb the PR still has a conflict to be resolved and unfortunately I'm not authorized to merge the fix. @lephyrus could you please merge it when the conflict is resolved?

lephyrus:
@bjorkblom I do not have write access and I'm not a maintainer of this project.
@njulhb You should resolve the conflict and reduce the changes to a single commit. Then we can try to get attention again.

dakolech:
Can we merge this one?

ocombe:
Done

bjorkblom:
Thanks @ocombe and GJ @njulhb

PR #663, opened 2017-09-10
docs(): update Ionic example

danielsogl:
I updated the Ionic example to our latest release and enabled lazy loaded pages by default.


ocombe:
Thanks!

PR #683, opened 2017-09-27
docs(README): Add plugin: MessageFormat Compiler

lephyrus:


ocombe:
thanks!

lephyrus:
@ocombe By the way, you may want to close https://github.com/ngx-translate/core/issues/150, even though it's hard to say what the exact expectation was there. But unless supporting more pluralization options is planned, I guess this is as good a time to close as any.

ocombe:
Good idea; thanks

Kranthi1547:
@lephyrus @ocombe Would you please update documentation/README providing examples to use pluralization with this merge

Issue #692, opened 2017-10-03
getBrowserLang throws "browserLang is undefined" exception in Firefox if no languages configured


dbaranovsky:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**
If remove languages from Firefox settings (so that all language related props in window.navigator are empty), then getBrowserLang method throws `browserLang is undefined` exception:
![image](https://user-images.githubusercontent.com/6683081/31151120-05b0f7aa-a88e-11e7-8e5c-a2c887552746.png)

![image](https://user-images.githubusercontent.com/6683081/31150955-80def7de-a88d-11e7-92b1-99e865c3b616.png)

**Expected/desired behavior**
`undefined` is returned

**What is the motivation / use case for changing the behavior?**
To handle language undefined case in consumer rather than have exception.

**Please tell us about your environment:**

* **ngx-translate version:** 7.0.0

* **Angular version:** 4.1.0
 
* **Browser:** Firefox

rbalet:
Close as it has been fixed for quite some time now: https://github.com/ngx-translate/core/blob/develop/projects/ngx-translate/src/lib/translate.service.ts#L572

PR #693, opened 2017-10-03
fix(browserLang) - #692: fix 'browserLang is undefined' error

dbaranovsky:


ocombe:
can you change it to "null" and change the return type of the function from `string` to `string | null` please? thanks

dbaranovsky:
@ocombe , done please check if that's what you expected

ffdybuster:
Now that `getBrowserLang()` returns a string or `null`, shouldn't `getBrowserCultureLang()` also return a string or `null` as well?

dbaranovsky:
`getBrowserCultureLang()` is not returning string or `null` to make it consistent

dbaranovsky:
@ocombe, @ffdybuster, could you please review the changes?

lib/src/translate.service.ts
```diff
@@ -560,9 +564,9 @@ export class TranslateService {
      *
      * @returns string
      */
-    public getBrowserCultureLang(): string {
+    public getBrowserCultureLang(): string | null {
```
ffdybuster: You should make sure that in case `browserCultureLang` is `undefined`, `null` is returned.
(I.e. do something like `return browserCultureLang || null;` at the end.)

dbaranovsky: fixed

dbaranovsky:
@ocombe , when do you think it can be merged?

patrickfav:
How is the status of this PR?

ocombe:
Closing this as another similar PR has been merged

Issue #662, opened 2017-09-08
empty translation should show original string


gemal:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[x] feature request
```

**Current behavior**
`"HELLO": ""`
shows blank i the view

**Expected/desired behavior**
I expect it to show `HELLO`
this way translators can see which strings they need to translate

**Reproduction of the problem**
If the current behavior is a bug or you can illustrate your feature request better with an example, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar. You can use this template as a starting point: http://plnkr.co/edit/tpl:WccVZSBM0rUgq2sXSUbe

**What is the expected behavior?**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

"@ngx-translate/core": "^7.1.0",
"@ngx-translate/http-loader": "^1.0.1",

* **Angular version:** 2.x.x
 
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

BlackZork:
A workaround for this is to exclude default language from extraction, i.e.:

```
    export class AppComponent {
      constructor(translate: TranslateService) {
        translate.setDefaultLang('none');
        translate.use('pl');
      }
    }
```
and in package.json:

`    "extract": "ngx-translate-extract --input ./src --output ./src/assets/i18n/{en,pl,de,fr}.json --clean --sort --format namespaced-json"`

This way you will see untranslated strings from source instead of empty replacements.

ffdybuster:
I don't think this very special use case should result in the default being changed. An empty string is a valid value and should *not* be treated the same as `null` (which is the correct value for "not translated").

You could write a parser plugin which replaces empty strings with `null`/`undefined`, or something like that.

IterationCorp:
I agree with @gemal this is completely killing the fallback feature!

I you don't want it as standard feature ok, but anyway we should have the choice to filter out empty string.

CodeAndWeb:
You can use this modified loader code. It removes empty strings from the files:

```
import {TranslateLoader} from '@ngx-translate/core';
import {HttpClient} from "@angular/common/http";
import 'rxjs/add/operator/map';

[...]

PR #698, opened 2017-10-09
Treat empty translation as missing

baso10:
Implementation for #662


SynovialWheat9011:
Expérience dans la traduction à continuer.

ocombe:
sorry but this would be a breaking change, and there are legitimate cases where you might want to use empty translations

PR #701, opened 2017-10-10
Provide your source language locale and don't load translations for t…

baso10:
Provide your source language locale and don't load translations for this language. It will also fail faster and provide key as translation


ocombe:
Hello and thanks for the PR, but I feel like this change is not needed anymore.

PR #709, opened 2017-10-23
merge translation object in default instead of replacing it

oicqcx:
the default value of  `shouldMerge` of `setTranslation()` is false, which makes it easy to overwrite the existed translation unexpectly, I think it is better to merge the translation objects than replace it.

### DIFF
--- a/src/translate.service.ts
+++ b/src/translate.service.ts
@@ -280,7 +280,7 @@ export class TranslateService {
      * @param translations
      * @param shouldMerge
      */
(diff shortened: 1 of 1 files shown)


keradus:
Changing the default value of public interface looks like a BC breaker

ocombe:
Hello, thanks for the PR but it would be a big breaking change, since it's easy to change the value if you need it, I'm closing this.

hussein-ctrl:
Hi everyone , I had a question related to @ngx-translate. Can we edit the translation files externally or add a translation file when we want to add a lang to our app.

PR #756, opened 2017-12-22
ngx-translate-zombies vscode extension

seveves:
I wrote a simple vscode extension for ngx-translate. When you have f.e. your english translation file ```en.json``` opened then you can hit ```F1``` and type ```ngxtz: find translation zombies in this file``` and hit return. The extension will search through all .ts, .js and .html files (excluding node_modules) and show a changed version of ```en.json``` in the vscode diff editor.

I needed this for my daily work and just wrote it ... so of course at the moment it only has very basic features but it's OSS and you can contribute if you want to :)


ocombe:
That's nice! thanks!

PR #765, opened 2018-01-08
Describe the latest compatible version with Angular 4

alexndreazevedo:
The information is missing on README.md. It's worth to be added.


ocombe:
Thanks

PR #767, opened 2018-01-09
fix(TranslateService): shouldMerge option now does merging with lodash

### LINKED ISSUE #764: Merge translations not working with ngx-translate-messageformat-compiler
```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] feature request
```

**Current behavior**

When using ngx-translate-messageformat-compiler, settings manually translations with 'shouldMerge' option will result in losing all translation values.

**Expected/desired behavior**

Should work the same as default compiler.

http://plnkr.co/edit/2qwydW

Merging translations using "setTranslation" method of TranslateService will cause removal of all translation values.

**What is the expected behavior?**

Be able to use messageformat's pluralization feature combined with lazy-loading translation.

* **ngx-translate version:** 9.0.2

* **Angular version:** 5.1.x
 
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

lihanyu23:
shouldMerge option of the `setTranslation` method didn't handle non-valid JSON translations object such compiled translations with custom compiler.
It is now doing deep merging with lodash.
Fixes #764

### DIFF
--- a/lib/src/util.ts
+++ b/lib/src/util.ts
@@ -61,8 +61,6 @@ export function isObject(item: any): boolean {
 }
 
 export function mergeDeep(target: any, source: any): any {
-    target = JSON.parse(JSON.stringify(target));
-    source = JSON.parse(JSON.stringify(source));
     let output = Object.assign({}, target);
     if (isObject(target) && isObject(source)) {
         Object.keys(source).forEach((key: any) => {
--- a/tests/translate.service.spec.ts
+++ b/tests/translate.service.spec.ts
@@ -183,6 +183,20 @@ describe('TranslateService', () => {
         });
     });
 
+    it("should merge non-valid JSON translations if option shouldMerge is true", () => {
+        translations = {};
+        translate.setTranslation('en', { "TEST": { "sub1": () => "value1" } }, true);
+        translate.setTranslation('en', { "TEST": { "sub2": () => "value2" } }, true);
+        translate.use('en');
+
+        translate.get('TEST.sub1').subscribe((res: string) => {
+          expect(res).toEqual('value1');
+        });
+        translate.get('TEST.sub2').subscribe((res: string) => {
+          expect(res).toEqual('value2');
+        });
+    });
+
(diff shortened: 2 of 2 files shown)


TakuyaAbe:
@lihanyu23 Thank you for [giving me a comment](https://github.com/ngx-translate/core/issues/764#issuecomment-356251932) & creating PR.

I find one point to fix.

You should add `lodash-es` as `dependencies`, not `devDependencies`.
If it is `devDependencies`, an error message as follows will appear when you install `@ngx-translate/core` & run server.

```
ERROR in /<PATH_TO_YOUR_ANGULAR_APPLICATION>/node_modules/@ngx-translate/core/src/util.ts (1,19): Cannot find module 'lodash-es/merge'.
```

Only `dependencies` are installed when we run `npm install @ngx-translate/core`.
Refer: https://stackoverflow.com/a/22004559/5977798

lihanyu23:
@TakuyaAbe thank you for pointing out !

I fixed that.

ocombe:
Hello, thanks for the PR but I don't want to add more dependencies to the library. If there's an issue with ngx-translate-messageformat-compiler it should be fixed in that library, not core.
I'm sorry to close this, since you put a lot of efforts into this, but I've got to makes choices.

lihanyu23:
@ocombe surely we can solve the issue by rewriting deepMerge function.

But it's not an ngx-translate-messageformat-compiler issue, the bug is in the core and relates to any compiler that compiles translation values into functions.

And if the function value as translation will still be supported, then this issue has to be solved here in the core.

ocombe:
Hmm ok, sounds fair. Can you do that then?

lihanyu23:
Just removed the JSON clean calls and keep the test.
Seems dirty but works.

ocombe:
Let's merge this, and see if people start to yell :D

PR #773, opened 2018-01-16
Change the translations setter variable to the correct one.

EddyNll:
The translations setter was pointing to currentLang instead of translations.


ocombe:
Hu, how did it stay like this for so long? Thanks!

82antu:
well done !!!

Issue #736, opened 2017-11-27
translation loader is not updating translations on change


rolandjitsu:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => check the FAQ and search github for a similar issue or PR before submitting
[x] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**
I'm using a custom loader, similar to [FIREBASE_EXAMPLE.md](https://github.com/ngx-translate/core/blob/master/FIREBASE_EXAMPLE.md), to get translations from Firebase. But it seems like if I change the db (from the FB dashboard), the changes are not reflected in the views.

**Expected/desired behavior**
The changes should be reflected when the translations observable emits.

**Reproduction of the problem**
If the current behavior is a bug or you can illustrate your feature request better with an example, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar. You can use this template as a starting point: http://plnkr.co/edit/tpl:XXwyUYS6ZL7qVD9I2l0g

**What is the expected behavior?**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **ngx-translate version:** 9.0.1

* **Angular version:** 5.0.3
 
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

alexndreazevedo:
It won't reflect the changes from the `Observable` `AngularFireDatabase.object()` because it only takes the first emitted value.

At [lib/src/translate.service.ts:267](https://github.com/ngx-translate/core/blob/master/lib/src/translate.service.ts#L267), the `TranslationService.getTranslation()` uses `take(1)` from `RxJS`.

Please, feel free to implement the solution and submit MR. Consider also a minimal [`debounceTime`](https://www.learnrxjs.io/operators/filtering/debouncetime.html) when implementing it.

kkotak:
This is important to fix as there are scenarios where you may wan to load a LocalStorage based default translation before retrieving the one from Firebase, which could have a visible delay. Also applies to PWA scenario.

MKhowaja:
> It won't reflect the changes from the `Observable` `AngularFireDatabase.object()` because it only takes the first emitted value.
> 
[...]

PR #777, opened 2018-01-22
Remved take(1) from getTranslation() to allow updating the stream mor…

kkotak:
Fix for - https://github.com/ngx-translate/core/issues/736

### DIFF
--- a/lib/src/translate.service.ts
+++ b/lib/src/translate.service.ts
@@ -264,7 +264,7 @@ export class TranslateService {
         this.pending = true;
         this.loadingTranslations = this.currentLoader.getTranslation(lang).pipe(share());
 
-        this.loadingTranslations.pipe()
+        this.loadingTranslations
           .subscribe((res: Object) => {
                 this.translations[lang] = this.compiler.compileTranslations(res, lang);
                 //Merge the new values for the translation with the old ones.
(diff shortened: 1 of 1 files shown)


lib/src/translate.service.ts
```diff
@@ -264,7 +264,7 @@ export class TranslateService {
         this.pending = true;
         this.loadingTranslations = this.currentLoader.getTranslation(lang).pipe(share());
 
-        this.loadingTranslations.pipe(take(1))
+        this.loadingTranslations.pipe()
```
alexndreazevedo: There is no need to the `pipe` if no operators are set.

ocombe:
If you remove the `take(1)` you'll run into memory leaks, and you'll have multiple loading translations observables in memory, triggering multiple times because there is no unsubscribe.

alexndreazevedo:
@ocombe, IMHO the loader should be the responsible for control when the translations should be loaded, not the core.

ocombe:
Agreed, but it would be a super breaking change, not sure how we could do that change smoothly

alexndreazevedo:
What about placing it as a setting disabled by default on `TranslateModuleConfig`? Additionally, the `ngx-translate/http-loader` could return an observable with `take(1)`.

Nikki1993:
Hello there o/ any update on the matter? Recently hit a similar scenario where I wish to pre-cache translations via LocalStorage to reduce the loading times but hit the snag where translations are not updated (by design).

MKhowaja:
With regards to this change being too breaking @ocombe I have a PR here with a different solution that should solve the same problem:
https://github.com/ngx-translate/core/pull/1028
Let me know what you guys think!

ocombe:
Closing this as the other PR is merged

PR #820, opened 2018-03-29
Remove outdated ionic example

CodeAndWeb:
The ionic example is outdated... still uses Angular 4. I think it makes more sense to isolate the example from the code. And it also seemed broken. Language switching did not work in my environment.

I've created an updated version here: https://github.com/CodeAndWeb/ionic-ngx-translate-demo
This one uses ngx-translate 9.1.1 and Angular 5


ocombe:
@danielsogl can you take a look at the new ionic example repository to check if everything works fine please? https://github.com/ngx-translate/ionic-example

danielsogl:
Sure the example is outdated. The example repository provided by @CodeAndWeb looks okay for me. 
@CodeAndWeb can you replace the old example with yours in the PR?

ocombe:
I think that keeping the example separated in another repository is better for now, it means that we can easily sync it with stackblitz and update it independently.
I might reorganize stuff once the angular cli v6 is released with official library monorepo support :)

Issue #547, opened 2017-05-28
getBrowserLang should be static


paulsouche:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[X] feature request
```

**Current behavior**

[getBrowserLang](https://github.com/ngx-translate/core/blob/master/src/translate.service.ts#L515) is not static

**Expected/desired behavior**
getBrowserLang should be static

**What is the motivation / use case for changing the behavior?**
Provide Browserlang without TranslateService deps

* **ngx-translate version:** latest

* **Angular version:** latest
 
* **Browser:** all

Thank you for the great job anyway !

SamVerschueren:
Why?

paulsouche:
Why ?

As you don't need a TranslateService to get browser language but a window and a navigator why instanciate it to get the method ? I'm not OK with you to say that would go passed the reason why we have dependency injection. Assume we have 2 instances of the class, this method is the same and does not depend of object state. That is obviously the definition of a static method...

How would you mock that in tests for instance ?

If getBrowserLang is static, it's only a function after all. You can stub it easily

```javascript
import { TranslateService } from '@ngx-translate/core';

// jasmine
spyOn(TranslateService, 'getBrowserLang').and.returnsValue('foo');

// sinon
sinon.stub(TranslateService, 'getBrowserLang');
```
Anyway this method should not be part of a class but module should export this function as a constant. It's an 'util' function...

ocombe:
That's true :)

PR #822, opened 2018-03-29
Make getBrowserLang() and getBrowserCultureLang() static

CodeAndWeb:
Issue #547

Should allow easier testing.


ocombe:
Hello, I understand the reasoning behind this, but as explained on the issue it's easy to stub it. Thanks for the PR in any case!

PR #871, opened 2018-06-14
add option to show keys and not translation

stefdelec:
- Allows users to see what is the key for a specific translation, in order to find quickly the value to change in .json
- My users have problem to find what key to change on our website. So they could add `window.desactivate_ngx_translate` in the console, and it would show keys and not translated values.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.pipe.ts
+++ b/projects/ngx-translate/core/src/lib/translate.pipe.ts
@@ -36,7 +36,7 @@ export class TranslatePipe implements PipeTransform, OnDestroy {
   }
 
   transform(query: string, ...args: any[]): any {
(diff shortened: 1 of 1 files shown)


lousley:
@stefdelec nothing...i have approved

ocombe:
hello, the idea is interesting, but you should add documentation for that, and a test if possible

balazsmaria:
why was this not merged?

daco:
I got this functionality working by first reset all languages, doing something like:
`availableLanguages.forEach(lang => this.translateService.resetLang(lang))`
And then set the new language to a non existing key, for instance:
`this.translateService.use("key")`. 
Now make sure that your translation loader returns an Observable of {}. Now you'll see all the translation keys on your website.

PR #881, opened 2018-06-27
Add option to merge translations fetched by the loader

Gaulomatic:
Update: I have updated the implementation due to the directive not working properly.

Explanation: This is useful for components shipping in a library which already provided translations via in-memory setTranslation() before theTranslateLoader had a chance to kick in. With this change you can provide TranslateModuleConfig.shouldMerge for the loader which in turn fetches the translations and merges them with the existing ones.


ocombe:
Closing this as it was fixed by another PR :)

PR #956, opened 2018-11-05
fix(TranslateService): compile translations only once

### LINKED ISSUE #955: Messages are compiled multiple times on initial load
## Current behavior
When an application is loaded initially the translations are compiled for every subscription to translateService.get on the current page because the translations are still pending when the observables are created.

The initial loading of the translations is triggered by calling translateService.setDefaultLang() in the app.component.ts. When translateService.get is called afterwards it re-compiles the translations in the subscription, because the initial loading is still pending (https://github.com/ngx-translate/core/blob/master/projects/ngx-translate/core/src/lib/translate.service.ts#L372).

## Expected behavior
Translations are only compiled once.

loadingTranslations could be set to an observable returning the compiled translations (maybe by piping a map in getTranslation(). Keep the return value of the method the same observable but set loadingTranslations to an observable with the piped map)

https://stackblitz.com/edit/github-kr4c1v
Every call to compile writes a message to the console

## Environment

<pre><code>
ngx-translate version: 10.0.2
Angular version: 6.1.7

Browser:
- All, independent of the browser and version
</code></pre>

saithis:
fixes #955

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.service.ts
+++ b/projects/ngx-translate/core/src/lib/translate.service.ts
@@ -238,18 +238,23 @@ export class TranslateService {
    */
   public getTranslation(lang: string): Observable<any> {
     this.pending = true;
-    this.loadingTranslations = this.currentLoader.getTranslation(lang).pipe(share());
+    const loadingTranslations = this.currentLoader.getTranslation(lang).pipe(share());
+    this.loadingTranslations = loadingTranslations.pipe(
+      take(1),
+      map((res: Object) => this.compiler.compileTranslations(res, lang)),
+      share()
+    );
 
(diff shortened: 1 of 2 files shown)


saithis:
@ocombe Can you please take a look at this? Because of this bug our app is unresponsive for ~5 seconds only compiling the messages over and over again when the initial page the user opens has many keys.

artuska:
@saithis This is an amazing pull request, thank you! This bug you've fixed ruined my Cordova web app — everything was fast in my browser on localhost but when i built my app and opened it on mobile phone I've been waiting ~15 seconds for my app to start :)

ocombe:
that's a great PR, thanks!

ocombe:
Available in 11.0.1

PR #972, opened 2018-12-13
feat(service): allow empty default language string

JuleSch:
An empty string is now also allowed as default language.
It checks if the default language is null. 
The check for 'undefined' is implied because of the use of == instead of ===.


JuleSch:
Hello,
is there something wrong with the execution or have I forgotten something else?
Otherwise our app will not be able to do the translations. :-/

SynovialWheat9011:
D' accord pour la traduction.

PR #973, opened 2018-12-13
docs(migration guide): add information about change to `prefix`

yitznewton:


yitznewton:
The change can be seen here:

https://github.com/ngx-translate/core/blob/v5.0.0/src/translate.service.ts#L86

vs.

https://github.com/ngx-translate/http-loader/blob/v0.0.2/src/http-loader.ts#L14

PR #990, opened 2019-01-17
Fix #707: Allow parameters in translation keys

### LINKED ISSUE #707: Parameters in translation keys
```
[x ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

Right now, if you use interpolation in the translation source/key you will get a translation is applied after the interpolation, so in the next example, if the version is 2.0, ngx-translate will look for a 'Version: 2.0' key in the file. 

```
<div translate> Version: {{ version }}</div>
```

The way this is currently supported after my understanding, is that you have to remove the parameters from the translation key/source like this:

```
<div [translateParams]="{ version: version }" translate > Version: </div>
```
The problem is that the translation key doesn't contain any information about the parameter names that the translator must provide, and when we extract and generate automatically the .pot template and give it to the translators, they have somehow to know the parameter names. Also it would be nice to make as little changes as possible in the angular templates when providing translations.

We thought of using some convention, for example like this:

```
<div [translateParams]="{ version: version }" translate > Version: [[version]]</div>
```

In this way the translators will know that the version is a parameter, and that they have to provide something like this; `{{version}}` in the translated text. 

I feel like this seems a little hack, and this is not the best way to work out this problem. Do you think there is a better way of doing this?  I imagine it is impossible to extract the 'original' uninterpreted value from the translate directive and use it as a key, instead of the interpolated value? I think this would be the easiest solution from my point of view....

Ben305:
Example: <button>{{ 'Log out ({ email })' | translate: { email: (user$ | async)?.email } }}</button>

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.parser.ts
+++ b/projects/ngx-translate/core/src/lib/translate.parser.ts
@@ -5,6 +5,7 @@ export abstract class TranslateParser {
   /**
    * Interpolates a string to replace parameters
    * "This is a {{ key }}" ==> "This is a value", with params = { key: "value" }
+   * "This is a { key }" ==> "This is a value", with params = { key: "value" }
    * @param expr
    * @param params
    */
@@ -21,7 +22,7 @@ export abstract class TranslateParser {
 
 @Injectable()
 export class TranslateDefaultParser extends TranslateParser {
-  templateMatcher: RegExp = /{{\s?([^{}\s]*)\s?}}/g;
+  templateMatcher: RegExp = /{{1,2}\s?([^{}\s]*)\s?}{1,2}/g;
 
   public interpolate(expr: string | Function, params?: any): string {
     let result: string;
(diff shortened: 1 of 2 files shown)


Ben305:
@biesbjerg Would you have a look?

Ben305:
@ocombe It would be nice if this regression could be fixed, has been working with previous versions

pavandaga:
Is there a workaround for this issue?

IvanoAlvino:
Hello everybody :)
Is this PR gonna be merged soon? Moving from angular-gettext to ngx-translate is going very smooth for me but it would be very nice to have this PR merged in order to having parameters in translations keys working as before.

Thanks in advance!

ocombe:
You should be able to do: ` {{ 'Log out ' + email | translate: { email: (user$ | async)?.email } }}`, I don't think this requires this change, because there's a good change that people actually use the symbols `{` and `}` in their text, and this would be a breaking change for them

PR #999, opened 2019-02-01
fix(TranslatePipe): subscriptions are assigned to EventEmitter props

trik:
make all subscriptions private, no need to expose them


ocombe:
I don't really see an issue in making those public, maybe people need to access them for some weird hack, and I don't mind them doing that :)

PR #1007, opened 2019-02-08
fix(TranslateService): check whether variable contains undefined value

cfilipponi:
When accessing a non-existing property, JavaScript returns undefined.
The check for 'undefined' will guarantee that browserLang has been and we can safely call .indexOf()


cfilipponi:
Another approach would be wrapping everything in a conditional like this:

```
if (browserLang) {
      if (browserLang.indexOf('-') !== -1) {
        browserLang = browserLang.split('-')[0];
      }

      if (browserLang.indexOf('_') !== -1) {
        browserLang = browserLang.split('_')[0];
      }
    }
```

Issue #998, opened 2019-01-31
Translate directive whitespace handling can cause erroneous recursive translation


dpmott:
## Current behavior
An HTML element with a translate directive and whose content is a translation key with leading/trailing spaces/newlines will return an erroneous recursive translation when the en.json file has an additional key matching the VALUE of the referenced translation key.

Example HTML:
```
<div translate>
  NAV_BAR.LINK_TEXT
</div>
```
Example en.json:
```
  "NAV_BAR": {
    "LINK_TEXT": "EXAMPLE"
  },
  "EXAMPLE": {
    "THIS": "this",
    "THAT": "that"
  }
```
With this example setup, the `<div>` will display `[object Object]`.

Note that the following HTML will _not_ trigger the bug:
`<div translate>NAV_BAR.LINK_TEXT</div>`

## Expected behavior
The `<div>` should display `EXAMPLE`.

## How do you think that we should fix this?
On this line: https://github.com/ngx-translate/core/blob/master/projects/ngx-translate/core/src/lib/translate.directive.ts#L82

Change `if (content!== node.currentValue)`
to
`if (trimmedContent !== node.currentValue)`
because 
https://github.com/ngx-translate/core/blob/master/projects/ngx-translate/core/src/lib/translate.directive.ts#L115
explicitly preserves whitespace in the node content, so only the trimmed content must be treated as the key.

## Minimal reproduction of the problem with instructions
https://github-ghfrtw.stackblitz.io

## Environment

<pre><code>
ngx-translate version: 10.0.2
Angular version: 6.1.10

This is still an issue in the latest ngx-translate version.
This was not an issue in ngx-translate 8.0.0

Browser:
- [X ] Chrome (desktop) Version 71.0.3578.98 (Official Build) (64-bit)
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

(Presumably other browsers as well; I didn't check them.)
 
</code></pre>

BADF00D:
I can confirm that this bug exists. I have a similar working example. In my version, the initial binding does work just fine (event with line breaks), but when I change the language at runtime, these values doesn't get updated.

[Render Stackblitz](https://ngxtranslate-whitespacebug.stackblitz.io)
[...]

PR #1016, opened 2019-02-19
fix(directive): avoid recursive errors when using keys with whitespaces

BADF00D:
The problem with issue #998 is, that at some node.originalContent in translate.directive.ts is overwritten by a translation that was assigned previously. So the translation key gets destroyed and no further translations could be made.

The bugfix introduces a new variable (node.lookupKey) that is generated once (by this.key or the trimmed content of the node) and used in further translations instead of calculating it over and over again. 

Since I changed no existing tests, I assume that alle existing functionality still works. I added new test to verify that the problem with white spaces are fixed.


ocombe:
Closing this as it was fixed by another PR (https://github.com/ngx-translate/core/commit/35427b071a177cca9e5b4e4c76e11a968559e9a7)

PR #1028, opened 2019-03-11
feat(service): add `getStreamOnTranslationChange` method

MKhowaja:
This is to fix the fact that the observable returned by the get method doesn't reactively update when translations change. We have an app that caches translations and makes a network call to retrieve more up to date translations but noticed that the observable doesn't emit the new translations.


jrolheiser-va:
Did `stream()` not handle your use case?

MKhowaja:
No it did not because its not the language that's changing but the translations when the backend call returns.

MKhowaja:
Alternatively we could include onTranslationChange in the concat of stream

XxAdi101xX:
We faced the same issue that @MKhowaja mentioned here and had to make some local fixes by using a wrapper functions to resolve it locally. This change would be super helpful.

ocombe:
Hello, could you add a test, and update the README with the docs please?

MKhowaja:
Yup!

MKhowaja:
@ocombe  While I was writing the tests, I realized that this method doesn't update if the lang changes. Do you think there is value in changing this method so it reactively updates onLangChange events as well?

PR #1066, opened 2019-05-07
docs(README): update shared modules link

cherurg:
The old link doesn't work anymore. So I've updated it.


ocombe:
Thanks a lot!

PR #1094, opened 2019-07-02
feat(core): upgrade dependencies for Angular v8+

peterblazejewicz:
We are using this lib on production and we migrated most of the content to version 8:

- dependencies upgrade
- minor changes after version upgrade
- minor configuration changes

Thanks!


projects/ngx-translate/core/src/lib/translate.pipe.ts
```diff
@@ -11,9 +12,9 @@ export class TranslatePipe implements PipeTransform, OnDestroy {
   value: string = '';
   lastKey: string;
   lastParams: any[];
-  onTranslationChange: EventEmitter<TranslationChangeEvent>;
-  onLangChange: EventEmitter<LangChangeEvent>;
-  onDefaultLangChange: EventEmitter<DefaultLangChangeEvent>;
+  onTranslationChange: Subscription;
```
juristr: Just curious: how comes this was changed from `EventEmitter` to `Subscription`?

peterblazejewicz: The `angular/core` team changed types for EventEmitter, relinquishing `any` and making these ones out-of-the-place:
https://github.com/angular/angular/commit/aebec4b1560063718a4d0cf1c42a8908668c0463#diff-c2e39457be0b071e3b8a08e84bc56583

DmitryEfimenko: I don't quite understand this either. The `any` type was updated to the `Subscription` type for the `.subscribe()` method of the `EventEmitter`. The `EventEmiter` type itself didn't change. I think this should have stayed unchanged

peterblazejewicz: I was under impression these are assignments for results of `subscribe` calls, later unsubscribed in the `_dispose`, not references for event emitter instances created to expose change state outside. I didn't go back to trace why the event emitter was used here in the first place. I understand purpose of the `EventEmitter` when used in the translate store and service, but here I'm not sure (what's the reason for me to subscribe here to detect changes in pipe state?)
This does not compile after upgrade (when core lib changes):
```text
Type 'Subscription' is missing the following properties from type 'EventEmitter<TranslationChangeEvent>': __isAsync, emit, subscribe, observers, and 16 more.
Type 'Subscription' is missing the following properties from type 'EventEmitter<LangChangeEvent>': __isAsync, emit, subscribe, observers, and 16 more.
Type 'Subscription' is missing the following properties from type 'EventEmitter<DefaultLangChangeEvent>': __isAsync, emit, subscribe, observers, and 16 more.
```
Simply put, this stops build fo the lib. This is a pipe and I don't see what's the case of using EventEmitter to expose pipe state change.

f-masche:
Is this pull request still active, and is there a possibility that this is merged anytime soon?
This is a big deal for many projects relying on `ng update` using angular 8.
Thanks you!

juristr:
> This is a big deal for many projects relying on ng update using angular 8.

Also Angular 9 being just about to land 😉. Tbh we upgraded our app to ng 8 a while back and we're using ngx translate. Didn't see any issues.

projects/ngx-translate/core/src/lib/translate.pipe.ts
ocombe: I can confirm that this change is necessary, you shouldn't use `EventEmitter` outside of `@Output`

PR #1105, opened 2019-07-26
feat(service): add option to set default language in root config

arunredhu:
- Default language option added for specifying the language at the time of defining the configuration in `forRoot` or `forChild` method. 

- Now user don't need to explicitly calls the `setRootLanguage` method for defining the default language at the initial time.

- Code formatted.

- Example added for `defaultLanguage`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,25 @@ export function createTranslateLoader(http: HttpClient) {
 export class AppModule { }
 ```
 
-#### 2. Init the `TranslateService` for your application:
+#### 2. Define the `default language` for the application
+
+```ts
+@NgModule({
+    imports: [
+        BrowserModule,
+        TranslateModule.forRoot({
+            defaultLanguage: 'en'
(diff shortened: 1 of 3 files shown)


ocombe:
hello and thanks for the PR, it's a good idea, but can you rebase it on master and remove the code formatting changes (it makes it hard to see the actual changes, and creates lots of conflicts)

arunredhu:
@ocombe  I have done the changes as you recommended. Please review and let me know in case you have any concerns.

BO41:
https://github.com/ngx-translate/core/pull/1105/files#diff-04c6e90faac2675aa89e2176d2eec7d8R242

Why set the default lang again if it was set in the root config?
Maybe add a comment to mention that not both is necessary?

btw. in markdown, if you use [only `1.` it will count up automatically](https://www.markdownguide.org/basic-syntax#ordered-lists) ;)

Issue #790, opened 2018-02-08
Issue with trailing whitespace in translation string when used with a directive


kyris9:
**I'm submitting a ...**
```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**
When the value of a key in the JSON file ends with a white-space, the language change does not have an effect **when used with a directive**. This works correctly when used with a pipe or with the service.

**Expected/desired behavior**
The value should get translated as defined in the JSON files.

**Reproduction of the problem**
**html**
`<span id="lblSearchCriteria" translate>common.terms.searchCriteriaLine</span>`
**en.json**
`{ "common": { "terms": { "searchCriteriaLine": "Search Criteria - " } } }`
**el.json**
{ "common": { "terms": { "searchCriteriaLine": "Κριτήρια Αναζήτησης - " } } }`
`

**Please tell us about your environment:**

* **ngx-translate version:** 9.1.1

* **Angular version:** 5.1.1
 
* **Browser:** [Chrome 64.0.3282.140]

CodeAndWeb:
Confirmed

https://stackblitz.com/edit/github-ad9uqm

liviu-c:
It seems this issue with the spaces (at the begining or at the end) is also available in the last build v11.0.0

https://stackblitz.com/edit/github-ad9uqm-vqauiz

liviu-c:
any news on this? the issue is valid in 11.0.1 as well.

Karl-Fokum-LogMeIn:
Any news on this issue?

Bat-Orshikh:
is there any on this issue?

jloupdef:
Not exactly the same problem as mine but might be resolved by my pending pull request #1117 as stated error is :
sh: 0: Can't open /etc/init.d/xvfb
The command "sh -e /etc/init.d/xvfb start" failed and exited with 127 during .

PR #1117, opened 2019-08-23
 Fix translationDirective ko after lang change

jloupdef:
When TranslateDirective is loaded before translations and translation is then updated through onTranslation callback, the key might be overridden with translated value and keep the value from being updated in case of future lang change. In the onTranslation callback, the node.currentValue is initialized directly from translation value and later compared from untrimmed node content (this.getContent(node)) which might contains line breaks and spaces. The fix consists in using only trimmed values to check for key update.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.directive.ts
+++ b/projects/ngx-translate/core/src/lib/translate.directive.ts
@@ -79,10 +79,10 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
           let trimmedContent = content.trim();
           if (trimmedContent.length) {
             // we want to use the content as a key, not the translation value
-            if (content !== node.currentValue) {
+            if (!node.currentValue || trimmedContent !== node.currentValue.trim()) {
               key = trimmedContent;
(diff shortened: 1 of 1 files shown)


jloupdef:
Travis CI build failed but not related to my change, there might be a problem with CI configuration as stated error is :
sh: 0: Can't open /etc/init.d/xvfb
The command "sh -e /etc/init.d/xvfb start" failed and exited with 127 during .

jloupdef:
Might also fix bug #790

isumeetk:
I also need this change urgently.  The issue with travis configuration can be fixed with this solution https://stackoverflow.com/questions/55674746/travis-sh-0-cant-open-etc-init-d-xvfb

libinemmanuel:
The trailing spaces causing some issue for me as well. Thanks for this fix. Eagerly waiting for the merge :)

ocombe:
I think that this should be fixed with https://github.com/ngx-translate/core/commit/35427b071a177cca9e5b4e4c76e11a968559e9a7, let me know if that's not the case

PR #1137, opened 2019-11-10
refactor: RxJS chains

martinsik:
I went through the code base and refactored some RxJS usage where I thought it could be improved or where it was using old and deprecated notations. I'm adding comments where it's not obvious what I did there.

In general there are two larger changes:
- Replacing deprecated `subscribe(v => void, func)` calls with `subscribe({ next: v => void, error: e => void})` https://github.com/ReactiveX/rxjs/issues/4159.
- Replacing Observable checks like `typeof res.subscribe === 'function'` with `isObservable(res.subscribe)`.

All existing tests were passing I just added one more additional test for some sync behavior.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.service.ts
+++ b/projects/ngx-translate/core/src/lib/translate.service.ts
@@ -1,5 +1,5 @@
 import {EventEmitter, Inject, Injectable, InjectionToken} from "@angular/core";
-import {concat, forkJoin, isObservable, Observable, Observer, of} from "rxjs";
+import {concat, forkJoin, isObservable, Observable, of} from "rxjs";
 import {concatMap, map, shareReplay, switchMap, take} from "rxjs/operators";
 import {MissingTranslationHandler, MissingTranslationHandlerParams} from "./missing-translation-handler";
 import {TranslateCompiler} from "./translate.compiler";
--- a/projects/ngx-translate/core/src/lib/translate.service.ts
+++ b/projects/ngx-translate/core/src/lib/translate.service.ts
@@ -1,6 +1,6 @@
 import {EventEmitter, Inject, Injectable, InjectionToken} from "@angular/core";
-import {concat, merge, Observable, Observer, of} from "rxjs";
-import {map, share, switchMap, take, toArray} from "rxjs/operators";
+import {concat, forkJoin, isObservable, Observable, Observer, of} from "rxjs";
+import {concatMap, map, shareReplay, switchMap, take} from "rxjs/operators";
 import {MissingTranslationHandler, MissingTranslationHandlerParams} from "./missing-translation-handler";
 import {TranslateCompiler} from "./translate.compiler";
 import {TranslateLoader} from "./translate.loader";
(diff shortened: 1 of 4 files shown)


projects/ngx-translate/core/src/lib/translate.service.ts
```diff
@@ -238,20 +238,27 @@ export class TranslateService {
...
+    const loadingTranslations = this.currentLoader.getTranslation(lang).pipe(
+      shareReplay(1),
       take(1),
+    );
+
+    this.loadingTranslations = loadingTranslations.pipe(
       map((res: Object) => this.compiler.compileTranslations(res, lang)),
-      share()
+      shareReplay(1),
+      take(1),
```
martinsik: This block is a little refactored because it was actually making multiple subscriptions when `this.currentLoader.getTranslation(lang)` was synchronous. In such case `share()` had no effect. There's a test https://github.com/ngx-translate/core/pull/1137/files#diff-3642e241860a02564009cf341f922874L396 that was supposed to test this scenario but it tests only async behavior. I added one more test down here https://github.com/ngx-translate/core/pull/1137/files#diff-3642e241860a02564009cf341f922874R410 that covers also sync `getTranslation()`

projects/ngx-translate/core/src/lib/translate.service.ts
```diff
@@ -308,22 +315,13 @@ export class TranslateService {
...
-        for (let k of key) {
-          let obs = typeof result[k].subscribe === "function" ? result[k] : of(result[k] as string);
-          if (typeof mergedObs === "undefined") {
-            mergedObs = obs;
-          } else {
-            mergedObs = merge(mergedObs, obs);
-          }
-        }
-        return mergedObs.pipe(
-          toArray(),
```
martinsik: This code is just refactored because it was unnecessarily complicated. Functionality should be the same.

projects/ngx-translate/core/src/lib/translate.service.ts
```diff
@@ -365,30 +363,15 @@ export class TranslateService {
...
     if (this.pending) {
-      return Observable.create((observer: Observer<string>) => {
-        let onComplete = (res: string) => {
-          observer.next(res);
-          observer.complete();
-        };
-        let onError = (err: any) => {
-          observer.error(err);
-        };
-        this.loadingTranslations.subscribe((res: any) => {
```
martinsik: Making subscriptions inside `new Observable.()` is almost never neccassy. It's also kind of a bad practise because it makes the outer Observable unsubscribable.
It seemed to me that this could be simplified significantly.

ocombe:
@martinsik it looks great! You're really better at rxjs than I am 😁
could you rebase the branch on master so that the tests can run please? if it's all green then I'll merge it!

iamnaumovnikita:
Guys, as I know the function `isObservable` should consume obseravble, but in your code it consumes subscribe function.
So, I think this code "isObservable(res.subscribe)" should be chnaged to "isObservable(res)" @ocombe and @martinsik please take a look

martinsik:
@iamnaumovnikita You mean this line? https://github.com/ngx-translate/core/pull/1137/files#diff-5a524f36ea5ac706856e90f772633f7f76a3a11e3a55e39e576045fc7c20f967R30

I think you're right, this is incorrect.

iamnaumovnikita:
I hve created PR https://github.com/ngx-translate/core/pull/1411 can you review it @martinsik @ocombe

Issue #874, opened 2018-06-22
TranslationService.reloadLang doesn't trigger onTranslationChange event


keradus:
Hi there!

I am using HttpLoader and I'm detecting that translations got modified on external URL, then I want to reload them.
Sadly, `reloadLang` doesn't work, as it fetches new dictionary indeed, but it doesn't trigger `onTranslationChange`, so all already printed translations are not updated.

As a workaround, I load translations manually and call `setTranslation` instead, which triggers `onTranslationChange`

juri33:
Same here

jonaslalin:
Same here

Odonno:
Same. But in our use case, it works perfectly.

rbalet:
Close as completed

PR #1139, opened 2019-11-12
docs(readme): add section on hot reloading of translations

Odonno:
I added a section in the FAQ to explain how to apply hot reloading of translations in an application. Thanks to the explanation of @keradus in #874


Odonno:
Until then...

PR #1142, opened 2019-11-15
fix(directive): trim white spaces to find matching keys

dpmott:
Issue #998: Translate directive whitespace handling can cause erroneous recursive translation


steg0139:
Nice work, looks like this has been an issue since version 10.0.0... LGTM

Langstra:
When is this getting merged and released?

PR #1169, opened 2020-02-12
[Issue 1163] Add additional tests

dpmott:
[Issue 1163] Add additional tests

EDIT: This also fixes issue https://github.com/ngx-translate/core/issues/1176

* Rolled back changes to translate.directive.ts from commit 24b7b2b8cb716adcb25905df9d64e5da43d41838 (but keeping tests)
* Added additional tests. 
* Issue 1163 appears to have actually been fixed by 1167/1164, as evidenced by the newly added tests.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.directive.ts
+++ b/projects/ngx-translate/core/src/lib/translate.directive.ts
@@ -72,28 +72,21 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
         if (forceUpdate) {
           node.lastKey = null;
         }
-        if(isDefined(node.lookupKey)) {
-          key = node.lookupKey;
-        } else if (this.key) {
+        if (this.key) {
           key = this.key;
         } else {
           let content = this.getContent(node);
           let trimmedContent = content.trim();
           if (trimmedContent.length) {
-            node.lookupKey = trimmedContent;
-            // we want to use the content as a key, not the translation value
-            if (content !== node.currentValue) {
-              key = trimmedContent;
-              // the content was changed from the user, we'll use it as a reference if needed
-              node.originalContent = content || node.originalContent;
-            } else if (node.originalContent) { // the content seems ok, but the lang has changed
+            if (node.originalContent && forceUpdate) { // the content seems ok, but the lang has changed
+              node.lastKey = null;
               // the current content is the translation, not the key, use the last real content as key
               key = node.originalContent.trim();
             } else if (content !== node.currentValue) {
               // we want to use the content as a key, not the translation value
               key = trimmedContent;
               // the content was changed from the user, we'll use it as a reference if needed
-              node.originalContent = content || node.originalContent;
+              node.originalContent = content;
             }
           }
         }
--- a/projects/ngx-translate/core/tests/translate.directive.spec.ts
+++ b/projects/ngx-translate/core/tests/translate.directive.spec.ts
@@ -1,4 +1,4 @@
-import {ChangeDetectionStrategy, Component, ElementRef, Injectable, ViewChild, ViewContainerRef} from '@angular/core';
+import {ChangeDetectionStrategy, ChangeDetectorRef, Component, ElementRef, Injectable, ViewChild} from '@angular/core';
 import {ComponentFixture, TestBed} from '@angular/core/testing';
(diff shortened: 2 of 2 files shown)


projects/ngx-translate/core/src/lib/translate.directive.ts
```diff
@@ -72,28 +72,21 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
         if (forceUpdate) {
           node.lastKey = null;
         }
-        if(isDefined(node.lookupKey)) {
-          key = node.lookupKey;
```
dpmott: I believe that the lookup key will prevent the translation of an updated interpolated bound template variable.

projects/ngx-translate/core/src/lib/translate.directive.ts
```diff
@@ -72,28 +72,21 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
...
+            if (node.originalContent && forceUpdate) { // the content seems ok, but the lang has changed
+              node.lastKey = null;
               // the current content is the translation, not the key, use the last real content as key
               key = node.originalContent.trim();
             } else if (content !== node.currentValue) {
               // we want to use the content as a key, not the translation value
               key = trimmedContent;
               // the content was changed from the user, we'll use it as a reference if needed
-              node.originalContent = content || node.originalContent;
+              node.originalContent = content;
```
dpmott: Because of the `if (trimmedContent.length)` conditional, `content` is guaranteed to be truthy, so the OR statement will never be executed.

projects/ngx-translate/core/tests/translate.directive.spec.ts
```diff
@@ -17,12 +17,14 @@ import {TranslateModule, TranslateService} from '../src/public_api';
     <div #leadingSpaceNoKeyNoParams translate> TEST</div>
     <div #trailingSpaceNoKeyNoParams translate>TEST </div>
     <div #withSpaceAndLineBreakNoKeyNoParams translate>
-      TEST
+      NESTED.TEST.KEY
+    </div>
+    <div #interpolatedWithSpaceAndLineBreakNoKeyNoParams translate>
+      {{ myInterpolatedTranslationKey }}
```
dpmott: Adding tests for interpolated bound template variables.

projects/ngx-translate/core/tests/translate.directive.spec.ts
```diff
@@ -51,7 +53,7 @@ describe('TranslateDirective', () => {
       ],
       declarations: [App]
     });
-    translate = TestBed.inject(TranslateService);
+    translate = TestBed.get(TranslateService);
```
dpmott: I saw that this had been recently changed, but the tests work either way, so I'm unsure why it changed?

If it needs to be `.inject()`, I'll happily put it back.

projects/ngx-translate/core/tests/translate.directive.spec.ts
```diff
@@ -164,19 +166,53 @@ describe('TranslateDirective', () => {
...
+  it('should update the DOM when the lang changes and the translation key is interpolated', async () => {
+    expect(fixture.componentInstance.interpolatedWithSpaceAndLineBreakNoKeyNoParams.nativeElement.innerHTML).toEqual(' Placeholder Text ');
+
+    const en1="This is a test - with trailing spaces in interpolated translation key";
+    const en2="Another test";
+    const fr1="C'est un test - pardon, je ne parle pas francais :)";
+    const fr2="Un autre test";
+    const setKey = key => {
+      fixture.componentInstance.myInterpolatedTranslationKey = key;
+      fixture.componentInstance.changeDetectorRef.markForCheck(); // Needed when changeDetection is ChangeDetectionStrategy.OnPush
```
dpmott: I'm not sure why the test harness is configured with `changeDetection: ChangeDetectionStrategy.OnPush`, as this isn't the default for Angular.  It's fine, but it also means that angular won't detect an interpolated variable change on its own, and we have to tell the change detector when to check for changes.

judomu:
We're having the same issue and so far I moved from directive to pipes to mitigate the issue, but it would be perfect to be able to use directive in all cases :)

tiaguinho (approved):

tiaguinho:
@ocombe can you take a look into this?

PR #1195, opened 2020-04-05
docs(README): added ngx-translate-testing

mwootendev:
Some utility classes I created a couple of years ago for helping test components with translations. Just updated for Angular 9 and @ngx-translate/core 12.0.0.


CodeAndWeb (approved):

CodeAndWeb:
Added it here: https://ngx-translate.org/resources/plugins/
That's all we can do currently...

PR #1093, opened 2019-07-02
Fix multi-line translates

anton-lebodkin:
Relates to this issue https://github.com/ngx-translate/core/issues/1092

PR #1203, opened 2020-04-29
Fix multi-line translates. Fixes #1092

### LINKED ISSUE #1092: Wrong behavior for translate directive in multi-line html nodes
## Current behavior
We have Prettier in our project which formats code this way. This kind of code makes changing translate language not working. Because there is extra space in the text node.

```
if (trimmedContent.length) {
                        // we want to use the content as a key, not the translation value
                        *if (content !== node.currentValue) {* <-- This line produce error
     because for example content === "Super account " and node.currentValue === "Super account", so this extra space makes it works wrong. Why instead of contect we don't pass trimmedContent???
                            key = trimmedContent;
                            // the content was changed from the user, we'll use it as a reference if needed
                            node.originalContent = this.getContent(node);
                        }
                        else if (node.originalContent && forceUpdate) { // the content seems ok, but the lang has changed
                            node.lastKey = null;
                            // the current content is the translation, not the key, use the last real content as key
                            key = node.originalContent.trim();
                        }
                    }
```

## Environment

<pre><code>
ngx-translate version: X.Y.Z
Angular version: X.Y.Z

Browser:
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

janpapenbrock:
Re-submit of #1093 

- [ ] Need to check this code and the tests to see if it is safe to merge.
- [ ] Also, evaluate why that branch of the `if-then-else` was duplicated, and see if there are additional errors

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.directive.ts
+++ b/projects/ngx-translate/core/src/lib/translate.directive.ts
@@ -82,7 +82,7 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
           if (trimmedContent.length) {
             node.lookupKey = trimmedContent;
             // we want to use the content as a key, not the translation value
-            if (content !== node.currentValue) {
+            if (trimmedContent !== node.currentValue) {
               key = trimmedContent;
               // the content was changed from the user, we'll use it as a reference if needed
               node.originalContent = content || node.originalContent;


janpapenbrock:
Apparently this has already been fixed in #1016 and I just don't see the results because we're on the 11.x release.

PR #1217, opened 2020-06-03
Updated type of TranslatePipe to properly represent already handled types

### LINKED ISSUE #794: Support AoT fullTemplateTypeCheck flag + AsyncPipe
```
[x ] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] feature request
```

**Current behavior**
Due to the type signature of the `TranslatePipe`, it doesn't currently place nice with Angular `fullTemplateTypeCheck` flag and `AsyncPipe`. The type signature of `AsyncPipe` is:

```ts
transform<T>(obj: null): null;
transform<T>(obj: undefined): undefined;
transform<T>(obj: Observable<T>): T | null;
transform<T>(obj: Promise<T>): T | null;
```

This means, in order to be compatible, the signature of `TranslatePipe` should change:

```ts
// Current
transform(query: string, ...args: any[]): any;

// New
transform(query: string | null | undefined, ...args: any[]): any;
```

At the very least, it should accept `null`. This seems like a pretty easy fix: Just don't call `updateValue` if `query == null` (double equals to handle `null | undefined`).

**Expected/desired behavior**
Translate pipe should be compatible w/ `AsyncPipe` and `fullTemplateTypeCheck`.

```ts
@Component({templateUrl: './foo.html'})
class Foo {
  public placeholder$: BehaviorSubject<string> = new BehaviorSubject('');
}
```
```html
<!-- foo.html -->
<input type="text" [placeholder]="$placeholder | async | translate" />
```
```js
// tsconfig.json AoT
{
  "angularCompilerOptions": {
    "fullTemplateTypeCheck": true,
  }
}
```
Results in:
```
Argument of type 'string | null' is not assignable to parameter of type 'string'.
  Type 'null' is not assignable to type 'string'.
```
Even though `$placeholder` will always be a string, `AsyncPipe` overload type is matching on `T | null`. While technically this could go back to Angular's type defs for `AsyncPipe`, there are likely reasons they are using a null union, and also good luck getting any movement on a real Angular issue.

NOTE: This is only going to affect certain things that require a string type. But for built-ins like `placeholder`, type widening is probably a non-starter.

Not having to wrap all components that use `$foo | async | translate`.

* **ngx-translate version:** ^8.0.0 - ^9.0.0

* **Angular version:** 5.2.4
 
* **Browser:** N/A

### LINKED ISSUE #1214: Missing types in TranslatePipe
Hi folks,

First of all let me thank you for the awesome work you've accomplished here! It's really been a pleasure to use.
Lately I updated an angular 9 project to perform extended checks when compiling.

```
# => tsconfig.json
...
"angularCompilerOptions": {
  "fullTemplateTypeCheck": true,
  "strictInjectionParameters": true
}
```

Unfortunately ever since I am getting annoying warnings, namely:
```
Argument of type 'string | undefined' is not assignable to parameter of type 'string' component.html:12:13 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.
  <span>{{possiblyUndefinedVariable | translate}}</span>
```

Of course I could wrap the entire block in an `*ngIf` but I feel that the pipes type should rather reflect that it may accept parameters of types other than `string`.
In fact, the specs and current implementation are defined in a way that the logic returns early, if it hits a falsy/empty value.

Implementation:
https://github.com/ngx-translate/core/blob/master/projects/ngx-translate/core/src/lib/translate.pipe.ts#L40-L43

Test:
https://github.com/ngx-translate/core/blob/master/projects/ngx-translate/core/tests/translate.pipe.spec.ts#L164-L171

Am I doing something complete off the charts here, or is my idea of defining the `query` parameter more accurate correct?
I would gladly provide a PR if this is the right direction.

PhilippMeissner:
This PR fixes https://github.com/ngx-translate/core/issues/794 and fixes https://github.com/ngx-translate/core/issues/1214.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.pipe.ts
+++ b/projects/ngx-translate/core/src/lib/translate.pipe.ts
@@ -37,7 +37,7 @@ export class TranslatePipe implements PipeTransform, OnDestroy {
     this.translate.get(key, interpolateParams).subscribe(onTranslation);
   }
 
-  transform(query: string, ...args: any[]): any {
+  transform(query: string | null | undefined, ...args: any[]): any {
     if (!query || !query.length) {
       return query;
     }


tiaguinho (approved):

PhilippMeissner:
@ocombe Maybe you can have a quick glance at this? It would really help us out to have this merged and released 👍

cice (approved):

KaiBirkenstock (approved):

Christop406 (approved):

PR #1224, opened 2020-06-22
Update TranslatePipe to working with Angular 9

Dok11:
I know it looks like hack but it works for now.
By some reasons `updateValue()` is called twice and at second time the `this.translate.get(key, interpolateParams)` returns the key of pipe instead of the real result from previous call.


Dok11:
Nevermind, it contains an error

PR #1232, opened 2020-07-12
fix(pipe): checking to observable

### LINKED ISSUE #1213: "[object Object]" if use translate pipe
https://github.com/ngx-translate/core/blob/c54a8342693f9a7cf53cc3b0626d6f5f4926b402/projects/ngx-translate/core/src/lib/translate.pipe.ts#L31

Should be `if (isObservable(res)) { `.

kirills98:
Fixing error when pipe return "[object Object]".

fix #1213


SEQUOiA87 (approved):

rbalet:
Already merged into the master

PR #1245, opened 2020-09-03
docs(README): add section for avoiding side-effects with HttpIntercep…

### LINKED ISSUE #921: Documentation: TranslateHttpLoader picks up HttpClient interceptors
## Current behavior
The documentation suggests using the TranslateHttpLoader for downloading the the translation files.  If there are `HttpInterceptors`, such as for Authentication, provided in the application, this can have unwanted side-effects.

## Expected behavior
Realistically, you only want to use the TranslateHttpLoader without HttpInterceptors.  This can be achieved using the HttpBackend class, as described [here](https://github.com/angular/angular/issues/20203#issuecomment-368680437).

Update the documentation to suggest this is a problem, and the alternative is to use the HttpBackend with HttpClient to ignore interceptors, like so:

```ts
export function translateHttpLoaderFactory(httpBackend: HttpBackend): TranslateHttpLoader {
    return new TranslateHttpLoader(new HttpClient(httpBackend));
}

...
        TranslateModule.forRoot({
            loader: {
                provide: TranslateLoader,
                deps: [HttpBackend],
                useFactory: translateHttpLoaderFactory
            }
        }),
```

Use an HttpInterceptor in the App that appends an external base url to all calls, thus breaking the location of the TranslateHttpLoader.

```ts
import {Inject, Injectable} from '@angular/core';
import {HttpEvent, HttpHandler, HttpInterceptor, HttpRequest} from '@angular/common/http';
import {Observable} from 'rxjs';

import {API_BASE_URL_TOKEN} from '../../shared/config.module';

@Injectable()
export class ApiBaseUrlInterceptor implements HttpInterceptor {
    private readonly apiBaseUrl: string;

    constructor(@Inject(API_BASE_URL_TOKEN) apiBaseUrl: string) {
        this.apiBaseUrl = apiBaseUrl;
    }

    public intercept(request: HttpRequest<any>,
                     next: HttpHandler): Observable<HttpEvent<any>> {
        return next.handle(request.clone({url: `${this.apiBaseUrl}${request.url}`}));
    }
}
```

## Environment

<pre><code>
ngx-translate version: 10.0.2
Angular version: 6.1.7

Browser:
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

blumk:
Resolves #921

Ideally `ngx-translate` would throw an exception if a translation file cannot be loaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,42 @@ export function createTranslateLoader(http: HttpClient) {
 export class AppModule { }
 ```
 
+##### Avoiding side-effects with HttpInterceptors
+
+If you are using custom `HttpInterceptors` (e.g. provided by an authentication library), you should bypass all `HttpInterceptors` when loading translation files.
+
+This can be achieved using the `HttpBackend` class, as described [here](https://github.com/angular/angular/issues/20203#issuecomment-368680437).
+
+```ts
+import {NgModule} from '@angular/core';
+import {BrowserModule} from '@angular/platform-browser';
+import {HttpBackend, HttpClientModule, HttpClient} from '@angular/common/http';
+import {TranslateModule, TranslateLoader} from '@ngx-translate/core';
+import {TranslateHttpLoader} from '@ngx-translate/http-loader';
+import {AppComponent} from './app';
+
+// AoT requires an exported function for factories
+export function HttpLoaderFactory(httpBackend: HttpBackend) {
+    return new TranslateHttpLoader(new HttpClient(httpBackend));
+}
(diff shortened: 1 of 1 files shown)


eclemens (approved):
It worked for me, i don't know why nobody has approve it

PR #1251, opened 2020-09-28
docs(README): reference light-localize-router

mohammedmatar:
we would like to share what we use in our projects need localized routes, we built another plugin for localized routes designed for angular 10 without the need of modifying angular router configurations.


rbalet:
Close in favor of : https://github.com/CodeAndWeb/ngx-translate.org/pull/5

PR #1268, opened 2020-11-19
 fix(directive): translate by attribute (not binding) with translation key as value

hosuaby:
Fix for angular 9 & 10

This PR fixes the behaiviour of `TranslateDirective` when it is used as a simple attribute `translate` (not attribute binding `[translate]`) with value and translated element already contains some text. Example:

```html
<div #attributeWithValue translate="TEST">Should be changed</div>
<div #attributeWithValueAndParams translate="TEST" [translateParams]="{value: 'great'}">Should be changed</div>
```

Currently (without this fix) this code will result in text translated twice. If value for `'TEST'` in english is "This is a test", after directive is executed, value of **#attributeWithValue** will be "This is a testThis is a test".

There was no such problem while using ng-translate with Angular 8. But since Angular 9 it appeared.

This PR also contains some refactorings.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.directive.ts
+++ b/projects/ngx-translate/core/src/lib/translate.directive.ts
@@ -3,6 +3,29 @@ import {Subscription, isObservable} from 'rxjs';
 import {DefaultLangChangeEvent, LangChangeEvent, TranslateService, TranslationChangeEvent} from './translate.service';
 import {equals, isDefined} from './util';
 
+function isTextNode(node: Node): boolean {
+  return node.nodeType === 3;
+}
+
+function getContent(node: any): string {
+  return isDefined(node.textContent) ? node.textContent : node.data;
+}
+
+function setContent(node: any, content: string): void {
+  if (isDefined(node.textContent)) {
+    node.textContent = content;
+  } else {
+    node.data = content;
+  }
+}
+
+interface NodeExt extends Node {
+  lookupKey: string;
+  lastKey: string;
+  currentValue: string;
+  originalContent: string;
+}
+
 @Directive({
   selector: '[translate],[ngx-translate]'
 })
@@ -17,14 +40,12 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
   @Input() set translate(key: string) {
     if (key) {
       this.key = key;
-      this.checkNodes();
     }
   }
 
   @Input() set translateParams(params: any) {
     if (!equals(this.currentParams, params)) {
       this.currentParams = params;
-      this.checkNodes(true);
     }
   }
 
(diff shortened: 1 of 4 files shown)


projects/ngx-translate/core/src/lib/translate.directive.ts
```diff
@@ -3,6 +3,29 @@ import {Subscription, isObservable} from 'rxjs';
...
+}
+
+function setContent(node: any, content: string): void {
+  if (isDefined(node.textContent)) {
+    node.textContent = content;
+  } else {
+    node.data = content;
+  }
+}
+
```
hosuaby, line 21: Interface for listing all custom fileds used by directive

projects/ngx-translate/core/src/lib/translate.directive.ts
```diff
@@ -17,14 +40,12 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
   @Input() set translate(key: string) {
     if (key) {
       this.key = key;
-      this.checkNodes();
```
hosuaby, line 20: Will be called by `ngAfterViewChecked`

projects/ngx-translate/core/src/lib/translate.directive.ts
```diff
@@ -17,14 +40,12 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
...
     }
   }
 
   @Input() set translateParams(params: any) {
     if (!equals(this.currentParams, params)) {
       this.currentParams = params;
-      this.checkNodes(true);
```
hosuaby, line 27: Will be called by `ngAfterViewChecked`

projects/ngx-translate/core/src/lib/translate.directive.ts
```diff
@@ -59,93 +80,87 @@ export class TranslateDirective implements AfterViewChecked, OnDestroy {
...
-            node.lookupKey = trimmedContent;
-            // we want to use the content as a key, not the translation value
-            if (content !== node.currentValue) {
-              key = trimmedContent;
-              // the content was changed from the user, we'll use it as a reference if needed
-              node.originalContent = content || node.originalContent;
-            } else if (node.originalContent) { // the content seems ok, but the lang has changed
-              // the current content is the translation, not the key, use the last real content as key
-              key = node.originalContent.trim();
-            } else if (content !== node.currentValue) {
```
hosuaby, line 92: This code block completelly diplucates the first one

projects/ngx-translate/core/src/lib/translate.pipe.ts
```diff
@@ -1,10 +1,8 @@
-import {ChangeDetectorRef, EventEmitter, Injectable, OnDestroy, Pipe, PipeTransform} from '@angular/core';
-import {isObservable} from 'rxjs';
-import {DefaultLangChangeEvent, LangChangeEvent, TranslateService, TranslationChangeEvent} from './translate.service';
+import {ChangeDetectorRef, OnDestroy, Pipe, PipeTransform} from '@angular/core';
+import {isObservable, Subscription} from 'rxjs';
+import {LangChangeEvent, TranslateService, TranslationChangeEvent} from './translate.service';
 import {equals, isDefined} from './util';
-import { Subscription } from 'rxjs';
 
-@Injectable()
```
hosuaby, line 7: Pipe no need to be injectable

projects/ngx-translate/core/tests/translate.directive.spec.ts
```diff
@@ -1,15 +1,16 @@
-import {ChangeDetectionStrategy, Component, ElementRef, Injectable, ViewChild, ViewContainerRef} from '@angular/core';
+import {ChangeDetectionStrategy, Component, ElementRef, ViewChild, ViewContainerRef} from '@angular/core';
 import {ComponentFixture, TestBed} from '@angular/core/testing';
 import {TranslateModule, TranslateService} from '../src/public_api';
 
-@Injectable()
```
hosuaby, line 5: Compoenent not need to be injectable.

PR #1275, opened 2020-12-29
fix(core): function overloading for instant

satanTime:
there's an issue if we use `no-any` rule which can be solved with function overloading.

```ts
const v1: string = translate.instant('test');
const v2: string = translate.instant(['test']); // should fail.
```


satanTime:
Hi @ocombe, Happy New Year! 

might you check the PR?

Thanks!

satanTime:
ping

satanTime:
Ping there

satanTime:
Hi @ocombe,

might you review when you have time?

Thank you in advance!

teunverhaert (approved):

Michael-Ziluck:
@tiaguinho This has been closed for awhile but it would be a much appreciated change. Do you know why this was closed when @teunverhaert approved?

PR #1279, opened 2021-01-06
fix(service): prevent `use()` from throwing an "Uncaught" exception

### LINKED ISSUE #1280: "Uncaught" exception is thrown when loader fails to load translations
When loader throws an exception, the subscription inside `TranslateService.use()` does not handle it, which leads to an "Uncaught" exception.
This exception leads to flaky unit tests, periodically failing with:
```
An error was thrown in afterAll
Uncaught [object Object] thrown
```

igodorogea:
Fixes #1280


rbalet:
Already merged

PR #1282, opened 2021-01-08
fix(directive): prevent infinite update for missing translation

Grolleau:


Grolleau:
Fixes #1280

zubSero (approved):
This should work. I tested it locally and it stopped the endless loop.

hthetiot:
This solve major performance issues we have, since we use Gettext approach and our english key are the same has our translation, the missing translation was triggered causing this infinit render.

korjikk:
Hi! Is there any specific reason this hasn't been merged yet?

rbalet:
If the problem still persiste, please provide a new PR

PR #1354, opened 2021-12-28
(feat) add method instantAll

simonegosetto:
added this method to allow users to get all values of current languages


CodeAndWeb:
That's not equivalent to what `instant()` does.
This one just returns the "raw" entries without processing them first. In cases where you use e.g. the message compiler, this would lead to very unexpected results because you would bet functions (!) instead of strings.

Btw. the implementation you are providing is already a the public api.

~~~
translateService.translations[translateService.currentLang]
~~~

Issue #602, opened 2017-07-13
Lazy loaded module specific JSON translation file not loading


tushamsbe:
**I'm submitting a ...**  (check one with "x")
```
[x ] feature request => load child module specific translation JSON file
```

**Current behavior**
In my lazy loaded child module `account.module.ts` I'm importing TranslateModule with forChild()
```
@NgModule({
  imports: [
    CommonModule,
    HttpModule,
    FormsModule,
    AccountRoutingModule,
    TranslateModule.forChild({
      loader: { provide: TranslateLoader, useClass: AccountTranslationLoader, deps: [Http] },
    })
  ],
})
export class AccountModule { }
```
I have custom translate loader `AccountTranslationLoader` written as 
```
export class AccountTranslationLoader implements TranslateLoader {

    constructor(private http: Http) {}

    getTranslation(lang: string): Observable<any> {
        return this.http.get(`./assets/i18n/account/${lang}.json`)
    }
}
```
Currently the expected json file request is never made and translation in child module views not working.

**Expected/desired behavior**
It is expected to load the child module json translation file and append it to the previously loaded translation file loaded from AppModule using TranslateModule.forRoot()

**Please tell us about your environment:**

* **ngx-translate version:** ^7.0.0

* **Angular version:** ^4.1.3
 
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

AlexFreem:
Subscribing.
Also expecting this problem, but with usage of factory based on TranslateHttpLoader.

matkarlg:
The GET request is never made because there are no subscribers on the Observable for fetching the child JSON. I assume translate.use() is used in the root to set language, which auto-subscribes behind the scenes [Source: translate.use](https://github.com/ngx-translate/core/blob/master/src/translate.service.ts#L221). It does not automatically fetch the JSON when the child is lazy loaded.

Make a test by subscribing to onLangChange() manually in the child module and change the language in code. It should fetch the JSON.

qifanrui:
I also encountered the same problem, who solved it?

qifanrui:
import { Http } from '@angular/http';
import { TranslateHttpLoader } from '@ngx-translate/http-loader';
import { NgModule } from '@angular/core';
import { IonicPageModule } from 'ionic-angular';
[...]

Issue #1266, opened 2020-11-13
Angular translation forchild not loading


Totot0:
I have an angular application that uses ngx-translation. For this, I made a Shared module that loads a translation file containing different keys (general errors...) :

`@NgModule({
  imports: [
    TranslateModule.forChild({
      loader: {
        provide: TranslateLoader,
        useFactory: TranslateFactory,
        deps: [HttpClient]
     }
    }),
  ],
  exports: [TranslateModule],
})

export class TranslatorModule {
  constructor(translate: TranslateService){
    translate.addLangs(['fr']);
    translate.setDefaultLang('fr');

    const browserLang = translate.getBrowserLang();
    translate.use(browserLang.match(/fr/) ? browserLang : 'fr');
  }
}

export function TranslateFactory(http: HttpClient) {
  return new TranslationService(http, 'assets/i18n');
}`

In my AppModule, I load this module and I also load another translation file via forRoot :

`@NgModule({
  declarations: [
    AppComponent,
  ],
  imports: [
    BrowserModule,
    AppRoutingModule,
    SharedModule,
    TranslateModule.forRoot({
        loader: {
          provide: TranslateLoader,
          useFactory: TranslateFactory,
          deps: [HttpClient]
       }
      }),
  ],
  bootstrap: [AppComponent]
})

export class AppModule{

}

export function TranslateFactory(http: HttpClient) {
  return new TranslationService(http, 'assets/i18n/myApp');
}`

My problem is that only translateService.forRoot works. Translations from translateService.forChild are not loaded

Then I just use my shareModule in my loaded modules in lazy loading mode

Yqnn:
I've the same issue, I think it's because the same TranslateService is shared in the 2 modules, and it uses the configuration given in 'TranslateModule.forRoot'.

docwhite:
For lazy-loaded modules with different translation loaders (loading `.json` from different files) it seems to be either (in the case of the lazy-loaded):

* (LazyModule `isolate: false`, `extend: true`) React to parent module translation events automatically without having to connect anything, just as they say, but cannot load the lazy loaded specific files.
* (LazyModule `isolate: true`, `extend: true`) We have to propagate changes to parent's translation event changes to the lazy child ourselves, and we can have our specific translations working! But the parent's translation won't work.

It's like I can't blend the two.

I got pretty close though maybe you could have a look and play within `StackBlitz`: https://stackblitz.com/edit/translations-and-lazy-loading?file=README.md

[...]

Issue #1193, opened 2020-04-03
Lazy-loaded modules dont get their translations when "isolate" is false.


thessoro:
## Current behavior

- Alternate i18n files are not loaded when using lazy-loaded modules if "isolate" param is false. So the module can access the main file translations but not theirs.

- When "isolate" is true the file is correctly loaded but the module doesn't have access to previously loaded translations.

## Expected behavior

Lazy loaded modules should be able to load their own translation files and at the same time being able to access previously loaded translation files as stated in the docs. 

## Minimal reproduction of the problem with instructions

For reproduction please follow the steps of the ngx-translate docs in a freshly angular created application with one or more lazy-loaded modules and one shared module exporting TranslateModule.

## Environment

<pre><code>
ngx-translate version: 12.1.2
Angular version: 9.1.0

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 10.15.2
- Platform:  Linux

Others:

</code></pre>

thessoro:
The workaround does not always work, for example if you get your language from store you could end with an untranslated component. It doesn't seem reliable for an enterprise-level solution.

phongca22:
#### app.module.ts
``` ts
export function createTranslateLoader(http: HttpClient) {
  return new TranslateHttpLoader(http, './assets/i18n/app/', '.json');
}

@NgModule({
  declarations: [AppComponent],
  imports: [   
    HttpClientModule,
    TranslateModule.forRoot({
      loader: { provide: TranslateLoader, useFactory: createTranslateLoader, deps: [HttpClient] }
    })
  ],
  providers: [],
  bootstrap: [AppComponent]
})
export class AppModule {}
```
#### app/en.json
``` json
{
   "action": "Create"
}
```
### Lazy loaded
order.module.ts
``` ts
export function createTranslateLoader(http: HttpClient) {
  return new TranslateHttpLoader(http, './assets/i18n/order/', '.json');
}

@NgModule({
  declarations: [OrderComponent],
  imports: [  
    TranslateModule.forChild({
      loader: { provide: TranslateLoader, useFactory: createTranslateLoader, deps: [HttpClient] },
      isolate: true
    })
  ]
})
[...]

PR #1361, opened 2022-01-19
fix(TranslateService): load translations if they have not been loaded before (#1193)(#1266)

dfmeretzki:
This PR contains no breaking changes.

When the `translateService.use(...)` method is called nothing happens if the current language is the same as the one passed as parameter. This causes that after changing the language in one instance of the translate service the rest of instances don't fetch the translations if they don't use isolate = true since the current language is the same as the one passed as parameter.  That is why workarounds like this
```
translateService.currentLang = '';
translateService.use ('en');
```
worked. Removing the first return statement leaves the service check if it has the translations and gets them if they have not been loaded before. It fixes (#1193)(#1266) between other issues I am not able to find now.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.service.ts
+++ b/projects/ngx-translate/core/src/lib/translate.service.ts
@@ -202,8 +202,9 @@ export class TranslateService {
    */
   public use(lang: string): Observable<any> {
     // don't change the language if the language given is already selected
+    let changeLang = true;
     if (lang === this.currentLang) {
-      return of(this.translations[lang]);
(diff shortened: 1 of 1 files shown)


dfmeretzki:
also fixes #602 #444  #425

Franky238 (approved):

Franky238 (approved):

Franky238 (approved):

liesahead:
Sad that library seems to be abandoned

rbalet:
Should be fixed in the latest version

PR #1371, opened 2022-05-13
bug fix: array with non-writable length

bahador-r:
When I use the setTranslation method in my project, I get the following error: 

ERROR Error: Uncaught (in promise): TypeError: can't define array index property past the end of an array with non-writable length
addLangs/<@http://localhost:4200/vendor.js:145411:20
addLangs@http://localhost:4200/vendor.js:145409:11

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.service.ts
+++ b/projects/ngx-translate/core/src/lib/translate.service.ts
@@ -302,7 +302,7 @@ export class TranslateService {
   public addLangs(langs: Array<string>): void {
     langs.forEach((lang: string) => {
       if (this.langs.indexOf(lang) === -1) {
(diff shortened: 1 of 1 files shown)


soheilrghb (approved):
This code does not logically change the project and prevents the error.

logicchemie (changes requested):
array with non-writable length

bellaabdelouahab (approved):

CodeAndWeb:
That change looks good for me.

However, how are you able to make this.langs (or this._langs) read-only from outside?

CodeAndWeb:
https://github.com/ngx-translate/core/commit/0818cec3107ea9bba24d9bf941201fd537380d98

Issue #216, opened 2016-08-30
Setting another value for a translation key should be updated


rabr2ro:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**

From what I could see, translations member, within TranslationService, for a specific language, is not a dictionary but an object tree. For example:

```javascript
translations[EN] = {
    'user': {
         'firstName': 'First Name',
         'lastName': 'Last Name'
    }
}
```
so, to use one of the keys in the pipe, you simply write {{ 'user.firstName' | translate }}
The signature for setting another value for one of the key is:

```javascript
public set(key: string, value: string, lang: string = this.currentLang): void
```

so, if one wants to change from code the value of the firstName, would inject TranslationService and do

```javascript
this.translationService.set('user.firstName', 'Only Name')
```
but the current implementation is not having the desired result, because, after executing the previous set, the translations object inside TranslatonService will look like this:

```javascript
translations[EN] = {
    'user': {
         'firstName': 'First Name',
         'lastName': 'Last Name'
    },
    'user.firstName': 'Only Name'
}
```

So, either I do not use correctly the set method, in which case the documentation should be updated, or it's internal implementation should be updated

```javascript
this.translations[lang][key] = value;
```

**Reproduction of the problem**
If the current behavior is a bug or you can illustrate your feature request better with an example, please provide the steps to reproduce and if possible a minimal demo of the problem via https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:btpW3l0jr5beJVjohy1Q).

**Please tell us about your environment:**

* **ng2-translate version:** 2.4.2

* **Angular version:** 2.0.0-rc.5
 
* **Browser:** [ Chrome XX | Firefox XX ] 
 
* **Language:** [ TypeScript 1.8 ]

ocombe:
Good point, I'll have to fix that :)

rabr2ro:
Thank you. I have a possible implementation for the problem, but it's in my own service that is a wrapper for the TranslationService. If you want, I can put here the code that I have (is TypeScript), you can take a look and you can refine it and included in the TranslationService implementation. Let me know if you would like to see what I have.

ocombe:
sure, that would help, thanks

rabr2ro:
Ok. So here it is:

```javascript
[...]

PR #1382, opened 2022-07-09
fix(service): allow setting translations in deep objects

hbsfaria:
This PR fixes the issue mentioned in #216 .

When we pass a key with nested properties, for example, `profilePage.header.name`, it will set the key as is instead of creating the necessary nested objects.

<table>
<tr>
<td> Currently </td> <td> After this PR </td>
</tr>
<tr>
<td>

```json
{
 "profilePage.header.name": "Test"
}
```

</td>
<td>

```json
{
 "profilePage": {
     "header": {
         "name": "Test"   
     }
  }
}
```

</td>
</tr>
</table>

This change will check if the received key has `.`, creating the nested objects automatically if they do not exist in the appropriate translation object.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.service.ts
+++ b/projects/ngx-translate/core/src/lib/translate.service.ts
@@ -458,7 +458,15 @@ export class TranslateService {
    * Sets the translated value of a key, after compiling it
    */
   public set(key: string, value: string, lang: string = this.currentLang): void {
-    this.translations[lang][key] = this.compiler.compile(value, lang);
+    const keys = key.split('.');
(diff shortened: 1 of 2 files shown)


mgorczynski2:
What is going on with this? Can anyone check it and merge it?

CodeAndWeb:
Integrated in upcoming V16
https://github.com/CodeAndWeb/ngx-translate/commit/02567229af93af6c915a73abd5b88ba41c4ed386

ocombe:
In your fork v16, please do not confuse people

CodeAndWeb:

PR #1399, opened 2022-12-26
feat: upgrade to angular 15

riderx:
upgrade and build test with angular 15


Oneilos (approved):
Thanks, i need it too

ocombe:
Why do you want to upgrade the project to v15? It should be working fine with v15 right now

riderx:
nom in strict mode don't allow to install without it

gaewynn (approved):

TiBz0u:
> npm in strict mode don't allow to install without it

Hi @riderx, even with this [peerDependency](https://github.com/ngx-translate/core/blob/5d04bd849876030b4117bf912a30c1a02a62df19/projects/ngx-translate/core/package.json#L22) that seems correct.
How do you trigger this error?

riderx:
install latest nodejs 16 or 18 and latest npm, then npm i this in a angular 15 project and you got it

1van0v (approved):

gaewynn (approved):

jberndsen:
We experience the same Angular 15 npm restriction. What would it take to get this PR merged?

matthiasferch:
Meanwhile, Angular 16 has been released and my team cannot upgrade because of our dependency on this plugin.

Is this repo still being maintained?

riderx:
that will be compatible to v15 as well ?

ocombe:
The current version is already compatible with v15... and I just tested, and it's also compatible with v16.
If you get an error about peer deps then you probably need to update npm because even with strict peer deps it works fine for me.

image

ocombe:
@matthiasferch what error do you get with angular v16 and current ngx translate ?

matthiasferch:
@ocombe 

```
Error: examples/showcase/src/app/app.module.ts:21:12 - error NG6002: 'TranslateModule' does not appear to be an NgModule class.

 21   imports: [
               ~
 22     CommonModule,
    ~~~~~~~~~~~~~~~~~
...
 32     AppRoutingModule,
    ~~~~~~~~~~~~~~~~~~~~~
 33   ],
    ~~~

  node_modules/@ngx-translate/core/public_api.d.ts:20:22
    20 export declare class TranslateModule {
                            ~~~~~~~~~~~~~~~
    This likely means that the library (@ngx-translate/core) which declares TranslateModule is not compatible with Angular Ivy. Check if a newer version of the library is available, and update if so. Also consider checking with the library's authors to see if the library is expected to be compatible with Ivy. 
```
    
I'm on npm 8.15.0 and node 16.17.0.

ocombe:
Since the current version has been built with v13 it is definitely compatible with ivy... usually this error happens when you install a library compatible with both ivy and the old view engine and you don't run ngcc on it, but since ngcc has been removed in v16, maybe it needs a new version of the lib in some cases ?

PR #1411, opened 2023-02-09
fix(pipe): isObservable usage

iamnaumovnikita:
### isObservable(res) instead of isObservable(res.subscribe)
I changed the usage of isObservable function to the correct one. Previously, it checks the subscribe function itself while it should check the observable itself. I also add a test that describes in which case it could lead to problem. 

### Without my fix the test is red and the actual result is 
`"{"operator": [Function anonymous], "source": {"_subscribe": [Function anonymous]}}" `

image

### With my fix the test is green.
image

In my test I described the scenario of a race condition when there is a bit delay of loading new language translation (after switching) and the usage MissingHandler with observable takes place.

### P.S.
In many cases the next tick will solve the problem and user will see the correct translation, but there is a chance to see the stringified observable `"{"operator": [Function anonymous], "source": {"_subscribe": [Function anonymous]}}" `

In my PR I also removed some redundat imports.

### DIFF
--- a/projects/ngx-translate/core/src/lib/translate.pipe.ts
+++ b/projects/ngx-translate/core/src/lib/translate.pipe.ts
@@ -28,7 +28,7 @@ export class TranslatePipe implements PipeTransform, OnDestroy {
     };
     if (translations) {
       let res = this.translate.getParsedResult(translations, key, interpolateParams);
-      if (isObservable(res.subscribe)) {
+      if (isObservable(res)) {
         res.subscribe(onTranslation);
       } else {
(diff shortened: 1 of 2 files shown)


CodeAndWeb:
Integrated in for ngx-translate v16
https://github.com/CodeAndWeb/ngx-translate/commit/e5558ff0c4b8cbf54184cf99603a37dd937e89d3

ocombe:
In your fork v16, please do not make people confused

CodeAndWeb:

PR #1428, opened 2023-06-14
Fix extend module load

rui-ad-monte:
On the constructor if is extended it will force load the default and the current language.
Possible fix for #1266


rbalet:
Closing this as the newest v17 logic should handle it properly

PR #1440, opened 2023-07-14
feat(core): Make TranslatePipe and TranslateDirective standalone

krtek4:
Make the TranslatePipe and TranslateDirective standalone so only those can be imported in a standalone component instead of the whole module.

### DIFF
--- a/packages/core/lib/translate.directive.ts
+++ b/packages/core/lib/translate.directive.ts
@@ -4,6 +4,7 @@ import {DefaultLangChangeEvent, LangChangeEvent, TranslateService, TranslationCh
 import {equals, isDefined} from './util';
 
 @Directive({
+  standalone: true,
   selector: '[translate],[ngx-translate]'
 })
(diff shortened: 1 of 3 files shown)


tarp-jusf:
Seeing as the Angular docs clearly state the following, there is no reason not to implement this change ASAP. It will literally not break anything.

`You can import a standalone component (or directive, or pipe) just like you would an [NgModule](https://angular.io/api/core/NgModule) - using [NgModule.imports](https://angular.io/api/core/NgModule#imports):`
Sourche: [Standalone Components - Using standalone components in NgModule-based applications](https://angular.io/guide/standalone-components#using-standalone-components-in-ngmodule-based-applications)

bjabeanjo (approved):

CodeAndWeb:
Included in https://github.com/CodeAndWeb/ngx-translate.

CodeAndWeb:

PR #1480, opened 2024-05-26
Fixes #1479: always translate to the last requested language

### LINKED ISSUE #1479: TranslateService.use with multiple calls causes to display the wrong language
## Current behavior
Changing the language multiple times in a short amount of time e.g. from the default language to a language which hasn't been loaded yet and then back to the default language (`"en"` -> `"de"` -> `"en"`) with the `TranslateService.use(...)` function causes the not yet loaded language (`"de"`) to still being used for texts instead of the last used language `"en"` (e.g. it's still `"Hallo Welt"` instead of `"Hello World"`).

We have encountered this problem because we have multiple sources where the language is set from (default langauge, browser language and from a user config saved/pulled from a server).

## Expected behavior
The last used language should be used in the texts (`"Hello World"` instead of `"Hallo Welt"`)

Stop the loading of the language when a new language is being set while loading

Use 2 translations (one as the default and another language which isn't loaded yet). Then create the following service:
```TypeScript
@Injectable({
  providedIn: "root",
})
export class SomeService {
  constructor(
    private translate: TranslateService,
  ) {
    this.translate.addLangs(["en", "de"]); // this is is normally a reference to a list of all our supported languages (generated dynamically)

    setTimeout(() => {
      this.translate.use("de"); // this language has not been loaded yet
      console.log(this.translate.currentLang); // "de"

      this.translate.use("en"); // default language
      console.log(this.translate.currentLang); // "en" - but all texts are still in "de"

    }, 10000); // wait a bit before testing
  }
}
```
My `app.module.ts` config for ngx-translate/core:
```TypeScript
...

export function httpLoaderFactory(http: HttpBackend) {
  return new MultiTranslateHttpLoader(http, ["./assets/common/i18n/", "./assets/i18n/"]);
}

@NgModule({
  ...
  imports: [
    ...
    TranslateModule.forRoot({
        loader: {
          provide: TranslateLoader,
          useFactory: httpLoaderFactory,
          deps: [HttpBackend],
        },
        compiler: {
          provide: TranslateCompiler,
          useClass: TranslateMessageFormatCompiler,
        },
        defaultLanguage: "en",
      }),
    ...
  ],
  ...
})
...
```

## Environment

<pre><code>
ngx-translate version: 15.0.0
Angular version: 17.3.5 and 16.2.14

Browser:
- [x] Chrome (desktop) version 124.0.6367.119
- [ ] Chrome (iOS) version XX
- [x] Firefox version 115.9.1esr
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [x] Edge version 124.0.2478.67
 
For Tooling issues:
- Node version: v20.12.2 
- Platform:  Windows

Others: Windows 10 22H2, Visual Studio Code, NPM
</code></pre>

El-Ahmed:
### The current behavior:
The language change event would fire whenever a new translation was loaded.
When requesting a translation, if another translation is loading, the system would wait for it to finish and return that instead of the translation of the most recent set language.

### This Pull Request addresses this by:
Fixing the issue mentioned in #1479.
Adding a test to ensure the fix works as expected.

### DIFF
--- a/packages/core/lib/translate.service.ts
+++ b/packages/core/lib/translate.service.ts
@@ -39,6 +39,7 @@ declare const window: Window;
 export class TranslateService {
   private loadingTranslations!: Observable<any>;
   private pending: boolean = false;
+  private lastLangUseRequest?: string;
   private _onTranslationChange: EventEmitter<TranslationChangeEvent> = new EventEmitter<TranslationChangeEvent>();
   private _onLangChange: EventEmitter<LangChangeEvent> = new EventEmitter<LangChangeEvent>();
   private _onDefaultLangChange: EventEmitter<DefaultLangChangeEvent> = new EventEmitter<DefaultLangChangeEvent>();
@@ -201,6 +202,9 @@ export class TranslateService {
    * Changes the lang currently used
    */
   public use(lang: string): Observable<any> {
+    // set the last requested language
+    this.lastLangUseRequest = lang
+
     // don't change the language if the language given is already selected
     if (lang === this.currentLang) {
       return of(this.translations[lang]);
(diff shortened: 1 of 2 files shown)


adessilly:
Is it going to be merged soon ? This fix would be very welcome ... Thx to @El-Ahmed

CodeAndWeb:

PR #1511, opened 2024-12-13
Fix: Check for null values in isDict method

Behrouz-B:
This pull request addresses a bug in the `isDict` method that previously did not explicitly check for null values. The original implementation of the `isDict` method only checked if the value was an object and not an array. 

This behavior is causing some translation values with null values to be converted into an empty object. 

**Changes made:**
1. Added a null check to the `isDict` method to ensure that null values are not incorrectly identified as dictionaries.

### DIFF
--- a/projects/ngx-translate/src/lib/util.ts
+++ b/projects/ngx-translate/src/lib/util.ts
@@ -56,7 +56,7 @@ export function isDefined(value: any): boolean {
 
 
 export function isDict(value: any): boolean {
-  return isObject(value) && !isArray(value);
+  return isObject(value) && !isArray(value) && value !== null;
 }
 
 


zargham-leanix:
@CodeAndWeb would you be able to review this change? This is blocker for us to update to v16. We have dynamic translations coming from backend, some of them are null values. There are being converted to empty objects now.

Behrouz-B:
> @CodeAndWeb would you be able to review this change? This is blocker for us to update to v16. We have dynamic translations coming from backend, some of them are null values. There are being converted to empty objects now.

@ocombe

CodeAndWeb:
I've seen the change but had no time yet.
Looks good. I must admit that I had not expected typeof 'null' to be 'object'...

CodeAndWeb (approved):

zargham-leanix:
@CodeAndWeb Andreas, would it be possible a minor release for this change?

CodeAndWeb:
Possibly tomorrow.

Can you take a look at this additional change?

https://github.com/ngx-translate/core/commit/d31e18bc0d6e63f8c06ba033b0f6a5830b4b377f

When passing `{a: {aa:"test", bb: "null}}`, you would get these values:

- `a` => `{aa:"test"}` the object without the null value
- `a.aa` => `test`
- `a.bb` => `a.bb` (the key)

so the `null` is removed from the results.

Behrouz-B:
> Possibly tomorrow.
> 
> Can you take a look at this additional change?
> 
> [d31e18b](https://github.com/ngx-translate/core/commit/d31e18bc0d6e63f8c06ba033b0f6a5830b4b377f)
> 
> When passing `{a: {aa:"test", bb: "null}}`, you would get these values:
> 
> * `a` => `{aa:"test"}` the object without the null value
> * `a.aa` => `test`
> * `a.bb` => `a.bb` (the key)
> 
> so the `null` is removed from the results.

Great! Looking forward to it. 
The additional changes look good and work well for us.

CodeAndWeb:
Done!

PR #1512, opened 2024-12-22
Fix getValue()

lmartelli:
It could not handle this case:
{
  "key1": {"key3": "value1"},
  "key1.key2": "value2"
}

I was not able to run the tests :-\

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 The internationalization (i18n) library for Angular.
 <br/>
 
-## Angular 16, 17, 18+
+## Angular 16, 17, 18, 19+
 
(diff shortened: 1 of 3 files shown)


CodeAndWeb:
Honestly, I don't think this case should be supported.

```
{
  "key1": {"key2": "value1"},
  "key1.key2": "value2"
}
```

Allowing this introduces ambiguity, as it's unclear which value should take precedence. This kind of structure can lead to confusion and potential bugs which are hard to spot.

I strongly prefer a clear, consistent structure that avoids such conflicts entirely.

PR #1517, opened 2025-01-14
task: deprecates EventEmitter and replaces it with Subject

### LINKED ISSUE #1513: Deprecate EventEmitter
# Current behavior
The library utilizes [EventEmitter](https://github.com/angular/angular/blob/main/packages/core/src/event_emitter.ts#L190-L194) to emit changes to the selected language.

---
# Expected behavior
The library should rely on a `Subject` instead, given that these are the most adequate to the use case.

## What is the motivation / use case for changing the behavior?
The `EventEmitter` is made to be utilized with components and directives, as per original angular team instruction.
Furthermore, with the start of moving from `@Output`/ `@Input` decorators to signals, `output`/ `input` respectively, the usage of the `EventEmitter` seems inadequate and possible to causes issues due to lack of support in the future.

## How do you think we should implement this?
Migrate the `onLangChange`, `onDefaultLangChange` and the `onTranslationChange` from an `EventEmitter`to a `Subject`.

I'm ok with forking the repo and doing the changes myself, however, I would like confirmation that is ok before putting work into it.

JsantosDK:
Closes #1513

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ To render them, simply use the `innerHTML` attribute with the pipe on any elemen
 
 - `currentLang`: The lang currently used
 - `currentLoader`: An instance of the loader currently used (static loader by default)
-- `onLangChange`: An EventEmitter to listen to lang change events. A `LangChangeEvent` is an object with the properties `lang: string` & `translations: any` (an object containing your translations).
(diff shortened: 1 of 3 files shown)


JsantosDK:
@CodeAndWeb PR should be done, and no issue that I can see with the replacement.
Could you please have a look and merge if all is ok?

CodeAndWeb:
Integrated using 5ebd7a58d8f36c150bd572e7c9f09e725d40dec6 - including some more refactoring.

PR #1527, opened 2025-02-20
feat: make interpolation functions protected to allow ineritance

de-don:
I have changed `private` to `protected` to have an ability to extend this class and override just one method. In my case I need to override logic in `interpolateString` method to use `''` instead of `substring` if value is not defined.

### DIFF
--- a/projects/ngx-translate/src/lib/translate.parser.ts
+++ b/projects/ngx-translate/src/lib/translate.parser.ts
@@ -36,12 +36,12 @@ export class TranslateDefaultParser extends TranslateParser
     return undefined;
   }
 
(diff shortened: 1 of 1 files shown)


CodeAndWeb:
4562aefbbd0b47757212ab72ff082d796ad38f9f

A more in-depth change for this. You now get even more flexibility :-)

de-don:
@CodeAndWeb Check please this comment https://github.com/ngx-translate/core/commit/4562aefbbd0b47757212ab72ff082d796ad38f9f#r152860552

I think there is a small bug, and I don't see any test about that logic :)

CodeAndWeb:
772cbd0b8c4d50bff165b7ba3c7c08a515ed613c

Fixed this. And also the not yet handled null.

PR #1529, opened 2025-03-07
Improve type safety of TranslateService and TranslatePipe

undsoft:
**Changes**

* `TranslatePipe` now explicitly accepts `null` and `undefined` and returns `Translation`. The code already guards against falsy values and specifying it explicitly helps when used with `strictNullChecks` in `tsconfig`. Otherwise, you may get a bunch of errors when an optional or nullable field is passed to the pipe.
* Replaced `any` with `unknown` in `InterpolationParams`.
* Fixes a test on non-English machines.

### DIFF
--- a/projects/ngx-translate/src/lib/translate.parser.spec.ts
+++ b/projects/ngx-translate/src/lib/translate.parser.spec.ts
@@ -29,7 +29,7 @@ describe('Parser', () => {
     });
 
     it('should support interpolation functions', () => {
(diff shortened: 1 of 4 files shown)


Christop406:
Would love this! Makes the pipe much easier to use in strict mode.

CodeAndWeb (commented):
Looks good to me.
I am a fan of stricter checks anyways... :-)

CodeAndWeb (approved):

Issue #1532, opened 2025-03-12
feat(service): use of signal


rbalet:
## Current behavior
We need to be working with subscription.

## Expected behavior
Although this is good the way it is, it could be nicer to add additional signal alternative.

## What is the motivation / use case for changing the behavior?
1. It would let us work with `computed` variable directly
2. It will lower the chances of forgotten unsubscription 

## How do you think that we should implement this?
_I can do a PR if needed_

I do see two options
### 1: Additional variable
Inside the `translate.service.ts`, add of additional signal for reach observable

```typescript
// Exemple

  $onTranslationChange = toSignal(this.store.onTranslationChange)
  get onTranslationChange(): Observable<TranslationChangeEvent> {
    return this.store.onTranslationChange;
  }
```

### 2: Replacement of the former logic 
_I'd avoid this as this is quite a breaking change_

Same as 1. but we remove the `onTranslationChange()` method. which let you write it as follow. 
```typescript
onTranslationChange = toSignal(this.store.onTranslationChange)
```

_I do personally always use the `$` in front of a signal, but standards aren't set yet for the nomenclature_

nxnen8282:
Our system attempted to address this issue:

Added signal alternatives for translation change events while maintaining existing Observable API. This allows working with computed variables directly and reduces chances of forgotten unsubscriptions.

Changes:
- Added imports for signal-related functionality
- Added signal alternatives with $ prefix for translation change events
- Maintained existing Observable-based API

You can examine the suggested modifications here: https://github.com/nxnen8282/ngx-translate-core-1532/commit/36bdf132967a597bf14521e494c9d3a60e7d1319.


> [!CAUTION]
> Advisory: This solution was generated using automated tools, and we strongly recommend reviewing the code thoroughly before implementation. The solution may be partial - please consider it as inspiration for your own implementation.

floAtEbcont:
@rbalet You only addressed `onTranslationChange`, but to be consistent, I think `get()` should have a signal equivalent too - here's a sample that works for me:

```typescript
translate(
        key: string | Signal<string>,
        params?: Record<string, string | number | null | undefined> | Signal<Record<string, string | number | null | undefined> | undefined>,
    ): Signal<string | undefined> {
        const key$ = toObservable(isSignal(key) ? key : signal(key));
        const params$ = toObservable(isSignal(params) ? params : signal(params));

[...]

PR #1550, opened 2025-05-18
feat(translate.service): $currentLang

rbalet:
## Description
Add `$currentLang`, which is reflecting `currantLang`.

## Note
Step into the use of signal for [the following issue](https://github.com/ngx-translate/core/issues/1532)


rbalet:
@CodeAndWeb Ci will be fixed by: https://github.com/ngx-translate/core/pull/1552

CodeAndWeb:
If made more complete by removing the old variables to avoid keeping the state in the signal and a variable in parallel.
4fc81c8182255dc1a13fdaae3666feb3c1523d4e

PR #1554, opened 2025-05-25
Translation transformer handler

victorlazzaroli:
Introduce a hook or extension point that allows modification of the raw translation value before the interpolation phase begins.

Issue: https://github.com/ngx-translate/core/issues/1555


CodeAndWeb:
We already have this functionality in form of the translate compiler.
It receives the translations before they are committed to the store.
The point in time is a bit earlier than your proposed hook.

victorlazzaroli:
I considered using the compiler to flatten the tree structure, but the cost is that it has to process the entire file every time it is loaded.
For this reason, I think it is more efficient and convenient to introduce the hook, especially for large files. 😊

CodeAndWeb:
Sorry for the delay...
Well - why don't you flatten your translation files before using them?
If they are delivered in a compressed form from your server, it should not make a big difference in size.
And you would spare all users to do the job that you can do once on the server.
Btw. converting the files is a one-click thing if you use a tool like https://www.codeandweb.com/babeledit

PR #1557, opened 2025-06-02
docs(README): browser caching warning

### LINKED ISSUE #1143: Suddenly, the translation system doesn't work for new words, old words translate but not new ???
Suddenly, the translation system doesn't work for new words, old words translate good but not new one ???

What can be problem ??

My settings in .json

"@ngx-translate/core": "^11.0.1",
"@ngx-translate/http-loader": "^4.0.0",

rbalet:
## Description 
Explain that browser caching may create a false sense of error and explain how to avoid it

Closes: https://github.com/ngx-translate/core/issues/1143

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,34 +41,42 @@ Simple example using ngx-translate:
 https://stackblitz.com/github/ngx-translate/example
 
 ### Table of Contents
-* [Installation](#installation)
-* [Usage](#usage)
-  * [Import the TranslateModule](#1-import-the-translatemodule)
-    * [SharedModule](#sharedmodule)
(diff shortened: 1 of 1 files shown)


CodeAndWeb (approved):

CodeAndWeb:
> Also, do we remove the README in favor of this documentation and then just add a link to it ?

This is my plan. I guess we could do that directly since the old versions of the docs are still available if somebody uses the tags for the versions they are looking for.

bdb43b7e0eaa1cfd9d1ec639841716bc89f70a96 ads a new configuration option that adds cache-busting automatically.

PR #1564, opened 2025-07-08
build(angular): v20.0.6

rbalet:
## Description
Also fixed the version to avoid having a wrong version being used. (removed `^` and `~`)


CodeAndWeb (approved):

CodeAndWeb:
Added with 2df8e2cc94e2ad9f8665514cb11dbd3ea1a72636

Issue #1566, opened 2025-07-09
feat(http-loader): handle multiple ressources & silent


rbalet:
## Current behavior
Actually, we can only target one ressource folder

## Expected behavior
Being able to select multiple one

## What is the motivation / use case for changing the behavior?
Remove the use of the external : https://www.npmjs.com/package/ngx-translate-multi-http-loader library for no additional cost

## How do you think that we should implement this?
The http-loader could accept, instead of prefix and suffix, a new resource and/or resources object.

```typescript
export interface TranslationResource {
  prefix: string
  suffix?: string
  optional?: boolean
}
```

## Optional
We could add an `optional` variable, which silent the HTTP call if there is an error
(It's also a long time wished issue)

CodeAndWeb:
Comments in the PR :-)

Viktor-Ivliev:
@CodeAndWeb We're really looking forward to version 17.1 🙏🏻 🙏🏻 🙏🏻

KacperKwietniewskieBilet:
@CodeAndWeb We're too 🚀

rbalet:
Closed as released in the v18 (doing some cleanup)

PR #1570, opened 2025-07-29
feat(http-loader): ressources

rbalet:
## Description
Enable loading multiple ressources with the default http-loader.  
Also add a `showLog` mode, which let use show, or not, the console.log error on missing file request

## Note
- It does accept `ressources: ['/i18n/first', '/i18n/second]` and `ressources: [prefix: { '/i18n/first'}]`
- If ressources is given, it does ignore the initial `suffix` and `prefix`.
- `showLog: boolean` shall be given, defaulted to false

### DIFF
--- a/projects/http-loader/src/lib/http-loader.ts
+++ b/projects/http-loader/src/lib/http-loader.ts
@@ -1,16 +1,18 @@
 import { HttpBackend, HttpClient } from "@angular/common/http";
 import { inject, Injectable, InjectionToken, Provider } from "@angular/core";
 import { mergeDeep, TranslateLoader, TranslationObject } from "@ngx-translate/core";
-import { forkJoin, map, Observable } from "rxjs";
+import { catchError, forkJoin, map, Observable, of } from "rxjs";
 
 export interface TranslateHttpLoaderResource {
     prefix: string;
     suffix?: string;
+    showLog?: boolean;
 }
 
 export interface TranslateHttpLoaderConfig {
     prefix: string;
     suffix: string;
+    showLog?: boolean;
     ressources: (string | TranslateHttpLoaderResource)[];
     enforceLoading: boolean;
     useHttpBackend: boolean;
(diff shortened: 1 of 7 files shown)


rbalet:
Closes: https://github.com/ngx-translate/core/issues/1566

projects/http-loader/src/lib/http-loader.ts
```diff
@@ -1,11 +1,19 @@
...
+export interface TranslateHttpLoaderResource {
+    prefix: string;
+    suffix?: string;
+    showLog?: boolean;
+}
 
 export interface TranslateHttpLoaderConfig {
     prefix: string;
     suffix: string;
+    showLog?: boolean;
```
CodeAndWeb, line 21: Adding logging is a good idea.

projects/http-loader/src/lib/http-loader.ts
```diff
@@ -39,9 +48,43 @@ export class TranslateHttpLoader implements TranslateLoader {
...
+                    }),
+                );
+            });
+
+            return forkJoin(requests).pipe(
+                map((response) => response.reduce((acc, curr) => mergeDeep(acc, curr), {})),
+            ) as Observable<TranslationObject>;
+        }
+
+        return this.http
```
CodeAndWeb: This is almost a duplication of the code above. I think we could handle this by always using the resources array instead.
If a single file is added, we can create resources array with 1 entry instead. 
Allowing both is only confusing. Who is going to win? obviously, when providing the resources array, the other one is ignored. but this is absolutely not transparent for the user. the resources loader also does not implement the cache-buster.


we could also use provideHttpLoader() and proviceMultiHttpLoader() or similar to make it clear what is happening here.

rbalet: @CodeAndWeb

1.  I would indeed only use the resource as this is a cleaner solution, but this would add a breaking changes. so I didn't wanted it '^^. I'll deprecate it for now

2. I'll refactor it right away

3. the resource loader does implement the cache-buster 
```typescript
return this.http.get<TranslationObject>(`${path}${cacheBuster}`).pipe(
```

4. I'm against adding a new provider, as it only complexify what could be done with a single one.

rbalet: @CodeAndWeb 

- I've corrected a `resources` typo
- I've deprecated the prefix & suffix. 
- I do let the showLog globally, as you could be willing to implement it only into a single child element -> Should I do the same for the cache burster. 

Let me know if it does fits your expectation and I'll propose a PR to the documentation

CodeAndWeb:
I've created 2 provider functions: One with the old interface to stay compatible, one with the new interface and the resources. 

Can you add some tests for loading 2 files?

rbalet:
Nice addition. 

I had to change the `provideTranslateMultiHttpLoader` fallback in case no resources where provided.

I've also implemented the tests, let me know what you think of it. 

The PR looks good to me

PR #1574, opened 2025-07-29
feat(service/use): sync html lang

### LINKED ISSUE #1573: feat(service): sync html lang
## Current behavior
Actually, the html `lang` attribute isn't reflected by the actually used lang. 

## Expected behavior
Automatically update his lang like follow

```html
<html lang="en">
```

## What is the motivation / use case for changing the behavior?
Better native SEO support, ease of use, and no negative impact for the library consumers

## How do you think that we should implement this?
Adding the following inside the `use()` method
```typescript
document?.documentElement?.setAttribute('lang', lang)
```

rbalet:
Closes: https://github.com/ngx-translate/core/issues/1573


CodeAndWeb:
https://ngx-translate.org/recipes/synchronize-lang-attribute-with-current-language/

Issue #884, opened 2018-07-02
 this.translate.set('HOME.TITLE') does not override nested JSON HOME.TITLE


FredUK:
## Current behavior
When setting a translation string with a nested key ie. translation.set('HOME.TITLE') a new key will be added to the translations instead of overriding the existing nested one

## Expected behavior
Should override existing translation

## How do you think that we should fix this?
Maybe the translations should automatically be flattened in memory when loaded ?

## Minimal reproduction of the problem with instructions
Reproduction project: https://github-xj88xp.stackblitz.io

Steps: Have a nested json translation with `"HOME": { "TITLE": "Hello world"}`
Call `this.translate.set('HOME.TITLE', 'New title', 'en');`
Note that getTranslation('en') now returns `"HOME": { "TITLE": "Hello world"}, "HOME.TITLE": "New title"` and string on template will not be overriden.

## Environment

<pre><code>
ngx-translate version:      "@ngx-translate/core": "^9.1.1" and (10.0.1 in stackblitz project above)
Angular version: "@angular/core": "^5.2.0" (and 6.0.0 in stackblitz project above)
Check whether this is still an issue in the most recent Angular version. Yes just checked stackblitz project
Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 6.9.2
- Platform:  Windows

Others:

</code></pre>

PedenFraHeden:
I just had the same issue today and ran through some of the code for "set" and "setTranslation" functions. And yes, "set" doesn't handle nested translation keys, unfortunately. But it's possible by using the "setTranslation" - but you have to create a nested object from your translation key manually. Here's what I did: 

` 
const key = "admin.users.Title";
const keyParts = key.split('.');
const last = keyParts.pop();
let translations = {};

keys.reduce(function (o, k) { return o[k] = o[k] || {}; }, translations)[last] = "New title";

this.translate.setTranslation("en", translations, true);
`
This will overwrite the value of admin.users.Title.

Chiragsorathiya:
how to set value of key ?
i have same problem

PedenFraHeden:
> how to set value of key ?
> i have same problem

[...]

Issue #758, opened 2017-12-24
Translateservice returns keys in service/Injectable constructor


romain10009:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => check the FAQ and search github for a similar issue or PR before submitting
[ ] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Current behavior**
using the `translateservice.get()` method inside a constructor of a provider/service in ionic 3 does return the translation Keys instead of their values in the subscription.

**Expected/desired behavior**
I would expect the `translateserice.get()` method to work in a service (@Injectable) constructor so that I can store the values retrieved in a plain javascript array to use them in my service's methods.

**Reproduction of the problem**

```
@Injectable()
export class UserService {

constructor(private translateService : TranslateService){

    private translations : [string];

    this.translateService.get(['SIGNUP_SUCCESS','DOES_NOTEXIST'])
      .subscribe(values=> {
        this.translations = values; //resolves to {DOES_NOTEXIST : "DOES_NOTEXIST", SIGNUP_SUCCESS:"SIGNUP_SUCCESS"}
      })
  }
}
```
In a @Page or outside of the constructor (inside a method of the UserService) it will work. My providers are added to the app.module.ts file. There the UserService provider is added to the providers Array. And there also is the following : 
```
TranslateModule.forRoot({
      loader: {
        provide: TranslateLoader,
        useFactory: (createTranslateLoader),
        deps: [HttpClient]
      }
    }),

```
**Please tell us about your environment:**

* **ngx-translate version:** 8.0.0

* **Angular version:** 5.0.0
 
* **Browser:** [Chrome 63]

alexndreazevedo:
Have you defined the default language or use any language?

https://github.com/ngx-translate/core/tree/v9.0.2#2-init-the-translateservice-for-your-application

romain10009:
Yes I do, in the constructor of my App Class.

```
export class MyApp {
  @ViewChild(Nav) nav: Nav;
  rootPage : any = null;

  constructor(private platform: Platform,
              private statusBar: StatusBar,
              private splashScreen: SplashScreen,
              private translate: TranslateService,
              private config : Config,
              private userService : UserService) {
    this.initializeApp();
    this.translate.setDefaultLang('fr');
  }
}
```

alexndreazevedo:
[...]

PR #1575, opened 2025-07-31
feat(store): initialized$

rbalet:
## Description
I've noticed that we aren't waiting the httpLoader before doing any data mutation, this does create the following error: 

- https://github.com/ngx-translate/core/issues/884
- https://github.com/ngx-translate/core/issues/758

## Do
We're now waiting for the initialized$ to be triggered, before letting using use the `set()` method.   
This later also has been transformed to a Promises, letting people be sure the value is set before moving forward (you can see this in the test).

## Note
We could surely fix other problem with this logic, but I'd prefer to do small PRs

### DIFF
--- a/projects/ngx-translate/src/lib/translate.service.ts
+++ b/projects/ngx-translate/src/lib/translate.service.ts
@@ -1,6 +1,6 @@
 import { inject, Injectable, InjectionToken } from "@angular/core";
-import { concat, defer, forkJoin, isObservable, Observable, of } from "rxjs";
-import { concatMap, map, shareReplay, switchMap, take } from "rxjs/operators";
+import { concat, defer, firstValueFrom, forkJoin, isObservable, Observable, of } from "rxjs";
+import { concatMap, filter, map, shareReplay, switchMap, take } from "rxjs/operators";
 import { MissingTranslationHandler } from "./missing-translation-handler";
 import { TranslateCompiler } from "./translate.compiler";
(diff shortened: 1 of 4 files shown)


CodeAndWeb:
I don't think that this is the problem in the ticket. The big happens long after loading. 
Users who want to set translations after loading some of them can use the subscription of `use()`.

But it might be indeed useful to have a `waitInitialized()` observable. 👍

I don't want to make `set()` async. This adds complexity that is not required - other functions like `instant()` also require an initialized service. 👎

The issue described here is more related to this: 
https://github.com/ngx-translate/core/commit/66964cf62480f3d8390ee97d7f10c1a22f9cabaf

rbalet:
@CodeAndWeb I was able to reproduce the error in the ticket and the PR does solve the problem.

Here's the step to reproduce

1. Inside the app.component.ts constructor, add a .set that override any value
2. Inside a timeout of 100ms, get this value 
3. You'll see, the value is not the one you've just .set, but the one from the file

Why?
Because the file needs to be loaded, as the .set is called almost directly
When the i18n file load, it then replace the .set value

I understand this PR to be adding a bit of complexity, but I believe the user won't be impacted (outside the fact that they won't be facing this bug anymore).

But maybe you have a better idea on how to solve this issue.

(Or maybe the step to reproduce is noir pointing to the error I thought it was '^^)

 Cheers

PR #1579, opened 2025-08-04
feat(service): hideBeforeDisplay

### LINKED ISSUE #689: Hide translation labels until language file is loaded
```
[x] support request => check the FAQ and search github for a similar issue before submitting
[ ] feature request
```

**Question**
How do I hide translation labels until language file is loaded? I assume this would involve some kind of cloak directive.

**Current behavior**
Problem is when I have something like `{{ Section.Title | translate }}`, `Section.Title` is visible until language file is loaded.

* **ngx-translate version:** latest
* **Angular version:** 4.0

rbalet:
## Description
Enable the possibility to hide the keys before they have been translated. 

- closed: https://github.com/ngx-translate/core/issues/689

### DIFF
--- a/projects/ngx-translate/src/lib/translate.pipe.ts
+++ b/projects/ngx-translate/src/lib/translate.pipe.ts
@@ -134,6 +134,10 @@ export class TranslatePipe implements PipeTransform, OnDestroy {
             });
         }
 
+        if (this.translate.hideBeforeDisplay && this.value === query) {
+            return "";
+        }
+
         return this.value;
     }
 
--- a/projects/ngx-translate/src/lib/translate.providers.ts
+++ b/projects/ngx-translate/src/lib/translate.providers.ts
@@ -28,6 +28,7 @@ export interface ChildTranslateServiceConfig extends Partial<TranslateProviders>
 export interface RootTranslateServiceConfig extends ChildTranslateServiceConfig {
     fallbackLang?: Language;
     lang?: Language;
+    hideBeforeDisplay?: boolean;
 
     /* @deprecated use `fallbackLang` */
     useDefaultLang?: boolean;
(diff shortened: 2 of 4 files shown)


rbalet:
@CodeAndWeb Note, I can also do the directive, just didn't wanted to add more time if you wouldn't accept this one or if you have a better implementation

CodeAndWeb:
I think this solution is incomplete: It does not work for - e.g. `instant()` or  directives.

A more complete solution might be to use a `MissingTranslationHandler` instead. E.g. we could create a 

~~~
export class SilentMissingTranslationHandler implements MissingTranslationHandler {
    handle(params: MissingTranslationHandlerParams): string {
        return "";
    }
}
~~~

and replace it with the "real" one after the first language or fallback language was loaded. I'd prefer this solution since it's more in one place instead of implementing the logic in several places.

Uses could also easily use a handler like this themselves.
It's also possible to register to `onLangChange()` and update the flag to show the translation keys.

rbalet:
@CodeAndWeb yes you're right and I like the idea :). Shall I let you implement it, as you're more familiar with the missing handler?

If not you can assign it to me and I'll look into it

CodeAndWeb:
It's already implemented and documented, see

https://ngx-translate.org/recipes/fix-translation-loading-glitches/#solution-3-custom-missingtranslationhandler-for-empty-strings

rbalet:
@CodeAndWeb Indeed, we already have workaround, but this implementation will let you delete the entire page, which is quite long, in favour of a simple `hideBeforeDisplay: true` settings. 

Or what is going against it ?

CodeAndWeb:
It does not replace the page. Options 1 and 2 are valid solutions to the problem, too - maybe even better than displaying an empty text - which event might introduce bigger content shift after the strings are loaded.
It puts more weight on our shoulders for a solution that can be solved with minimum effort.

What I would accept is adding a `useEmptyStringIfNotFound` or similar as parameter for the `MissingTranslationHandler`.
That has the same effect and keeps the changes to a minimum.

Issue #782, opened 2018-01-26
translate.instant with empty string throws error: Parameter "key" required


ghiscoding:
**I'm submitting a ...**  
```
[x] bug report
```

**Current behavior**
Doing a `translate.instant('')` on an empty string throws an error of `Parameter "key" required` but doing the same on a white space doesn't throw an error. This is thrown by the [instant()](https://github.com/ngx-translate/core/blob/master/lib/src/translate.service.ts#L462) function itself
```javascript
public instant(key: string | Array<string>, interpolateParams?: Object): string | any {
        if(!isDefined(key) || !key.length) {
            throw new Error(`Parameter "key" required`);
        }
```

**Expected/desired behavior**
Don't throw error on empty string, I often initialize my form elements with empty string and wish to have the translation working. Now because of this issue I need to check if it's not empty string then translate else it throws a load of errors. 

**Reproduction of the problem**
Just put this anywhere in your code: `console.log(this.translate.instant(''))`

**What is the expected behavior?**
No error in console

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**
Windows 10, Angular 5

* **ngx-translate version:** 9.1.1

* **Angular version:** 5.0.0
 
* **Browser:** all

cookieksv:
I have the same issue.
Would be perfect if the "instant" method was simply returning the same provided to it empty value (or undefined, or null, etc).

tibistibi:
have the same error, it breaks the page and nothing is translated. on my page it happens sometimes, i guess when it loads to fast or not fast enough. although i have a check that the object is present.

ghiscoding:
The only thing I found to bypass this issue was to pass a white space, so if the translation key is incorrect then instead use a white space. Something along this
```ts
const myKey = '';
this.translate.instant(myKey || ' '); // this will use the white space since myKey is null or empty
```

tibistibi:
thanks, that will work but i rather stupid.... why is this issue not take care of :disappointed:

boisamazing:
@ghiscoding 

Thanks for your workaround, and it solves the same error with _translate.stream_ where it is parsing an empty array. So a white space in array is required:

`const myArray = [" "];`
`this.translate.stream(myArray);`

QuickDirty, but it did save lots of time. Have to say that this issue is worth to be taken care. 🚶

exedor:
[...]

PR #1580, opened 2025-08-04
feat(service/instant, get): accept empty string

rbalet:
## Description
avoid throwing an error on wrong value, as it breaks more than if fix. 

Solve: https://github.com/ngx-translate/core/issues/782


CodeAndWeb:
I am not sure if they wanted to create it this way. Returning an empty string instead of a translation.
We could provide a translation for an empty key - e.g. in a file like this:

~~~
{
      "": "this is empty"
}
~~~

But here it get's a bit strange: `get(".")` could also work like this:

~~~
{
      "": {
         "" : "another"
      }
}
~~~

rbalet:
I'm pretty sure the implementation satisfy the issue -> https://github.com/ngx-translate/core/issues/782#issuecomment-370376844

PR #1584, opened 2025-08-05
feat(service): getTranslations

### LINKED ISSUE #1583: Missing access to `translationService.translations`
In `@ngx-translate/core` 16 we were doing this:

```ts
translateService.translations[language]
// gives access to a Record with all translations for the language, e.g:
// {
//   "a": "b",
//  ...
// }
```

In `@ngx-translate/core` 17 I couldn't find similar API to access "all translations for the language".
I see the private `translationService.store` property, but it's typescript-private.

How can we access "all translations for the language" now?

rbalet:
## Description
Let us get the translations object from the store

Closes: https://github.com/ngx-translate/core/issues/1583

### DIFF
--- a/projects/ngx-translate/src/lib/translate.service.ts
+++ b/projects/ngx-translate/src/lib/translate.service.ts
@@ -120,6 +120,7 @@ export abstract class ITranslateService {
         key: string | string[],
         interpolateParams?: InterpolationParameters,
     ): Observable<Translation>;
+    public abstract getTranslations(language: Language): InterpolatableTranslationObject;
 
     public abstract setTranslation(
         lang: Language,
@@ -523,6 +524,13 @@ export class TranslateService implements ITranslateService {
         }
 
         return makeObservable(this.getParsedResult(key, interpolateParams));
+    }
+
+    /**
+     * @returns The translations for the given language
+     */
+    public getTranslations(language: Language) {
+        return this.store.getTranslations(language);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


projects/ngx-translate/src/lib/translate.service.ts
```diff
@@ -120,6 +120,7 @@ export abstract class ITranslateService {
         key: string | string[],
         interpolateParams?: InterpolationParameters,
     ): Observable<Translation>;
+    public abstract getTranslations(language: Language): InterpolatableTranslationObject;
```
CodeAndWeb: ```DeepReadonly<InterpolatableTranslationObject>```

projects/ngx-translate/src/lib/translate.service.ts
```diff
@@ -525,6 +526,13 @@ export class TranslateService implements ITranslateService {
         return makeObservable(this.getParsedResult(key, interpolateParams));
     }
 
+    /**
+     * @returns The translations for the given language
+     */
+    public getTranslations(language: Language) {
```
CodeAndWeb: ```DeepReadonly<InterpolatableTranslationObject>```

CodeAndWeb:
We should use ```DeepReadonly<InterpolatableTranslationObject>``` as return type to prevent users from changing translations.

projects/ngx-translate/src/lib/translate.service.ts
rbalet: Done

projects/ngx-translate/src/lib/translate.service.ts
rbalet: @CodeAndWeb I did this, but I'm not sure I had to, as the method was inferring his type from the store. 

But both are good for me ; )

kirillgroshkov:
What is a next step to get this merged?

@CodeAndWeb @rbalet

rbalet:
@CodeAndWeb This PR should be safe, would you find time to merge it?

PR #1589, opened 2025-08-10
feat(service): created hasTranslation method that checks if translation key exists

### LINKED ISSUE #1128: Add check for presense of a translation
## Current behavior
There is no way to be certain about the existence of a translation for a specific key. The best thing I can do is checking if the translation is the same as the given key:
`if (this.translate.instant(myKey) === myKey) { ... }`
However, besides not being pretty, this cannot detect the case when the translation really *is* the same as the key.

## Expected behavior
It would be nice to have a safe and concise test for the presence of a translation.

## How do you think that we should implement this?
`if(this.translate.hasTranslation(key)) { ... }`

Nice bonus: Give it an optional argument for testing another language instead of the current:
`hasTranslation(translationKey: string, language?: string): boolean`

kamkom:
### Description
Let us check if translation key in given language exists

closes: https://github.com/ngx-translate/core/issues/1128

### DIFF
--- a/projects/ngx-translate/src/lib/translate.service.ts
+++ b/projects/ngx-translate/src/lib/translate.service.ts
@@ -6,7 +6,7 @@ import { TranslateCompiler } from "./translate.compiler";
 import { TranslateLoader } from "./translate.loader";
 import { InterpolateFunction, TranslateParser } from "./translate.parser";
 import { TranslateStore } from "./translate.store";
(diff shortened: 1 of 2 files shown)


CodeAndWeb:
I see that we could do that - but what's the use for it?

See https://github.com/ngx-translate/core/issues/1128#issuecomment-3182818742

I am happy to add new features - but I also want to keep the library as lean as possible and not add unnecessary functionality or things that can be achieved in other ways.

Issue #1590, opened 2025-08-11
Add possibility to inform on fallback language use


HerrDerb:
The language store can be configured with a fallback language in case a translation is missing in the current target language. This is very practical as it avoids that a ugly translation key is displayed to the user.

## Issue
When configuring the store with a fallback service and a specific token was missed to translate, the ui will show the fallback translation.
This happens without any notification what so ever. By having no pointer that a translation token is actually not translated, the error might not be recognized and fixed, probably presenting a confusing translation of the fallback language to the user.
Currently, there is no way to know when a fallback translation is being used, which makes it hard to catch missing translations and ensure proper localization.

## Implementation idea
By adding a new subject to the store (`onFallbackLangUse`) which triggers on fallback lang use,  it would allow consumers to subscribe to these notifications and take appropriate action (e.g., logging, reporting, analytics, etc.).

vladsadretdinov:
@HerrDerb maybe i am wrong - but library has the **MissingTranslationHandler API**, where you can put your logic/logger/handler for situation when application faces with missing translation key.

PR #1591, opened 2025-08-11
feat(translation-store): add subject to inform on fallback lang use

HerrDerb:
Adds https://github.com/ngx-translate/core/issues/1590 (Add possibility to inform on fallback language use)

### DIFF
--- a/projects/ngx-translate/src/lib/translate.service.ts
+++ b/projects/ngx-translate/src/lib/translate.service.ts
@@ -65,6 +65,10 @@ export interface FallbackLangChangeEvent {
     translations: InterpolatableTranslationObject;
 }
 
+export interface FallbackLangUseEvent {
+    lang: string;
+    missingTranslationKey: string;
+}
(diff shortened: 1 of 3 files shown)


rbalet:
Thanks for this PR. 

Shall I know what you put it as `private` and how is your plan of consuming it?

HerrDerb:
@rbalet sorry, the getter was missing 🙈

HerrDerb:
and tests are still missing too

rbalet (approved):

CodeAndWeb:
Thank you for the PR.

I’m afraid I can’t accept it at this time. It would add complexity to the library while offering limited benefit to most users.

We already provide similar functionality through the [MissingTranslationHandler](https://ngx-translate.org/reference/missing-translation-handler-api/), which you can easily customize in your project to achieve the same result.

HerrDerb:
@CodeAndWeb must have overlooked that one (y)

PR #1596, opened 2025-08-20
feat(store): loaders

### LINKED ISSUE #1585: fix(provideChildTranslateService): not calling the loader when the lazy loaded module is first loaded
Hi,

not sure if its just me doing something wrong, I have a module federated application that I worked with and recently updated to V17 to test things out.

out main/host application uses a custom loader and missing translation handler which is provided like so:
```typescript
        provideTranslateService({
          missingTranslationHandler: provideMissingTranslationHandler(MissingTranslationHandler),
          loader: provideTranslateLoader(HostTranslationLoader)
        }),
```

Within our remote application that is lazy loaded later I have the following provided:
```typescript
        provideChildTranslateService({
          extend: true,
          missingTranslationHandler: provideMissingTranslationHandler(MissingTranslationHandler),
          loader: provideTranslateLoader(RemoteTranslationLoader)
        })
```
The loader for the remote is slightly different in order to load the translation assets from a custom path that is hosted separatly.

The the host application first loads the host translation loader gets to work and loads all of the translations for the host no problem.

when i navigate to the remote application I can see that the child translate service is instantiated and has the correct lang an such set, however, this does not trigger the initial loading of the remote translations. This is also the same when the application changes language. The host loader kicks in and gets the new translations but the remote one does nothing...

Am i missing an initial bootstrap / event registration or do I need to do this manually in each remote to work?

Edit:

I can also add if in the constructor/init function of my remote / lazy loaded application and call this.translateService.use('en') I can see that the RemoteTranslationLoader does load the translations but it feels odd that I have to manually trigger this...

1. Open the test-app
2. Open your chrome devTool
3. Go to the extended page
4. Change the lang
5. Only the extended loader will be fired

rbalet:
## Description
As we're now having multiple instance of the `TranslateService`, the only way to sync language across all services is through the `store`.

As [this issue](https://github.com/ngx-translate/core/issues/1585) describe, when adding a new loader, and changing the lang, the parent loader isn't triggered anymore. 

## Cause
The loading happen inside the service, but this service is not aware of the other loaders

## Proposed solution
Storing each `loader` inside the `store`, and calling it when needed. 

- New `loaders` map inside the `store` class
- On service instantiation, add the new loader into the `loaders` map
- On service destroy, remove this loader (language aren't updated as this does not seems required)
- On language change, triggering every loaders.

## Food for though
This PR is more "food for though" than a finite PR, as I'm aware following have not been solved
- Should we empty the translation on destroy? (I think we should not)
- Only the last compiler will be triggered, (we could also store compilers into a map and this would solve the error)
- This triggers one huge request, and we have to await it, maybe an other way could be optimized.

### DIFF
--- a/projects/ngx-translate/src/lib/translate.service.ts
+++ b/projects/ngx-translate/src/lib/translate.service.ts
@@ -1,4 +1,4 @@
-import { inject, Injectable, InjectionToken } from "@angular/core";
+import { inject, Injectable, InjectionToken, OnDestroy } from "@angular/core";
 import { concat, defer, forkJoin, isObservable, Observable, of } from "rxjs";
 import { concatMap, map, shareReplay, switchMap, take } from "rxjs/operators";
 import { MissingTranslationHandler } from "./missing-translation-handler";
@@ -6,7 +6,7 @@ import { TranslateCompiler } from "./translate.compiler";
 import { TranslateLoader } from "./translate.loader";
 import { InterpolateFunction, TranslateParser } from "./translate.parser";
 import { TranslateStore } from "./translate.store";
-import { insertValue, isArray, isDefinedAndNotNull, isDict, isString } from "./util";
+import { insertValue, isArray, isDefinedAndNotNull, isDict, isString, mergeDeep } from "./util";
 
 /**
  * Configuration object for the translation service.
@@ -173,7 +173,7 @@ export abstract class ITranslateService {
 }
 
 @Injectable()
-export class TranslateService implements ITranslateService {
+export class TranslateService implements ITranslateService, OnDestroy {
     private loadingTranslations!: Observable<InterpolatableTranslationObject>;
     private pending = false;
     private _translationRequests: Record<Language, Observable<TranslationObject>> = {};
@@ -184,6 +184,8 @@ export class TranslateService implements ITranslateService {
     private parser = inject(TranslateParser);
     private missingTranslationHandler = inject(MissingTranslationHandler);
     private store: TranslateStore = inject(TranslateStore);
+
+    private _loaderIndex!: number;
 
     private readonly extend: boolean = false;
 
@@ -225,6 +227,8 @@ export class TranslateService implements ITranslateService {
     }
 
     constructor() {
+        this._addLoader();
+
         const config: TranslateServiceConfig = {
             extend: false,
             fallbackLang: null,
@@ -245,6 +249,17 @@ export class TranslateService implements ITranslateService {
         if (config.extend) {
             this.extend = true;
         }
+    }
+
+    private _addLoader() {
+        while (this.store.loaders.has(this._loaderIndex)) {
+            this._loaderIndex++;
+        }
+        this.store.loaders.set(this._loaderIndex, this.currentLoader);
+    }
+
+    ngOnDestroy() {
+        this.store.loaders.delete(this._loaderIndex);
     }
 
     /**
@@ -341,9 +356,22 @@ export class TranslateService implements ITranslateService {
     ): Observable<InterpolatableTranslationObject> {
         this.pending = true;
 
-        const loadingTranslations = this.currentLoader
-            .getTranslation(lang)
-            .pipe(shareReplay(1), take(1));
+        const loaders = Array.from(this.store.loaders.values());
+        const requests = loaders.map((loader) => loader.getTranslation(lang).pipe(take(1)));
+
+        // Merge all translation objects
+        const loadingTranslations = (
+            requests.length > 1
+                ? forkJoin(requests).pipe(
+                      map((results: TranslationObject[]) =>
+                          results.reduce(
+                              (acc, curr) => mergeDeep(acc, curr),
+                              {} as TranslationObject,
+                          ),
+                      ),
+                  )
+                : requests[0]
+        ).pipe(shareReplay(1), take(1));
 
         this.loadingTranslations = loadingTranslations.pipe(
             map((res: TranslationObject) => this.compiler.compileTranslations(res, lang)),
--- a/projects/ngx-translate/src/lib/translate.service.ts
+++ b/projects/ngx-translate/src/lib/translate.service.ts
@@ -179,8 +179,8 @@ export class TranslateService implements ITranslateService, OnDestroy {
     private _translationRequests: Record<Language, Observable<TranslationObject>> = {};
     private lastUseLanguage: Language | null = null;
 
-    public currentLoader = inject(TranslateLoader);
-    public compiler = inject(TranslateCompiler);
+    protected currentLoader = inject(TranslateLoader);
+    protected compiler = inject(TranslateCompiler);
     private parser = inject(TranslateParser);
     private missingTranslationHandler = inject(MissingTranslationHandler);
     private store: TranslateStore = inject(TranslateStore);
--- a/projects/ngx-translate/src/lib/translate.service.ts
+++ b/projects/ngx-translate/src/lib/translate.service.ts
@@ -185,8 +185,6 @@ export class TranslateService implements ITranslateService, OnDestroy {
     private missingTranslationHandler = inject(MissingTranslationHandler);
     private store: TranslateStore = inject(TranslateStore);
 
-    private _loaderIndex!: number;
-
     private readonly extend: boolean = false;
 
     /**
@@ -227,8 +225,6 @@ export class TranslateService implements ITranslateService, OnDestroy {
     }
 
     constructor() {
-        this._loaderIndex = this.store.addLoader(inject(TranslateLoader));
-
         const config: TranslateServiceConfig = {
             extend: false,
             fallbackLang: null,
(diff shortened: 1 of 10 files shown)


projects/ngx-translate/src/lib/translate.store.ts
CodeAndWeb, line 1: I’d suggest using an array of objects like `{loader: TranslateLoader, count: number}` to track the loaders.

The store could then provide a few helper methods:

* `addLoader(loader: TranslateLoader)` adds a new entry if the loader doesn’t exist yet, or increases the count if it does. This prevents duplicate entries.
* `removeLoader(loader: TranslateLoader)` decreases the count, and removes the entry completely if the count reaches 0.
* `getLoaders()` returns a unique list of loaders, so the same loader isn’t triggered multiple times when calling `getTranslation()`.

This way, loader management stays tidy and avoids duplication.

projects/ngx-translate/src/lib/translate.service.ts
```diff
@@ -247,6 +251,17 @@ export class TranslateService implements ITranslateService {
         }
     }
 
+    private _addLoader() {
```
CodeAndWeb: I feel like the responsibility for handling the loaders (add/remove) would fit better in the store rather than the service.  
The service already has quite a lot on its plate, and my goal is to reduce its responsibilities instead of adding more.  

Something along the lines of:  

- `this.store.addLoader(this.currentLoader)`  
- `this.store.removeLoader(this.currentLoader)`  
- `this.store.getLoaders()`  

---

We could also remove the `currentLoader` variable, since there’s no longer a single “current” loader. Keeping it would just add confusion.

Instead, the calls could look like this:

* `this.store.addLoader(inject(TranslateLoader))`
* `this.store.removeLoader(inject(TranslateLoader))`

This keeps things clear and aligns better with the new loader management approach.

---

Another thing to watch out for: each new `TranslateService` instance currently adds a loader to the list.  
If the injection hierarchy doesn’t override the loader, the same loader could end up being added multiple times.  
This would cause the same translation file to be loaded repeatedly. I've proposed a solution for this in the loader comment.

CodeAndWeb (changes requested):

projects/ngx-translate/src/lib/translate.service.ts
rbalet: 1. Responsibility of the loader 
Yes, but what if a user would like to add/remove a loader manually?   
For what I understand, the store shouldn't be accessed directly, and therefore it could make sense to have it inside the service. 

Indeed it's starting to be a bit complex, but this is also because we propose a lot of features. 
But you do you on this topic, as you know it better than me

----
2. `currentLoader`
Would be nice to get rid of it, yes

----
3. `TranslateService`
The constructor is called only if you `provide` the service, therefore it is not possible to add multiple time the same loader without wanting it. 

If the user call multiple time the same "endpoint" then it's his own fault (but maybe he'd like to have a different language loader in different part of his app, therefore I would keep it as is)

projects/ngx-translate/src/lib/translate.store.ts
rbalet: @CodeAndWeb Edit: 

The logic implemented does, as far as I know, avoid implementing duplicate loaders, but only if the user decide to duplicate it (by providing a new loader with the same endpoint).

I wouldn't add the `count`. 

as I've decided to loop through the map object, find the next slot available, and add in there the new loader. 

### Why ? 
Let's say you 

A -> add his loader -> index 0 [A]
B -> add his loader -> index 1 [A, B]
C -> Add his loader -> index 2 [A, B, C]
B -> remove his loader ->  [A, undefined, C]
D -> Add his loader -> [A, D, C]
E -> Add his loader -> [A, D, C, E]

Doing so I'm avoiding loosing space in the map
(Let me know if it makes sense '^^)

CodeAndWeb: If you use this:

~~~
provideTranslateService({
  loader: provideTranslateLoader(CustomLoader),
  compiler: provideTranslateCompiler(CustomCompiler)
})
~~~

and use a child without adding a loader:
~~~
  provideChildTranslateService({
    // no loader
  })
~~~

it would lead to adding the same loader twice. 

Your solution prevents removing it from the list completely if one of the services calls OnDestroy.
But it does not prevent triggering the load functionality twice.

projects/ngx-translate/src/lib/translate.store.ts
```diff
@@ -31,6 +31,31 @@ export class TranslateStore {
 
     loaders = new Map<number, TranslateLoader>();
```
CodeAndWeb: make `loaders` private. use the functions to access it.

projects/ngx-translate/src/lib/translate.store.ts
```diff
@@ -31,6 +31,31 @@ export class TranslateStore {
...
+
+    removeLoader(loaderIndex: number) {
+        if (this.loaders.has(loaderIndex)) this.loaders.delete(loaderIndex);
+    }
+
+    getLoaders() {
+        return this.loaders;
+    }
+
+    getLoader(loaderIndex: number) {
```
CodeAndWeb: getLoader(loader) is obsolete - it's used nowhere.

projects/ngx-translate/src/lib/translate.store.ts
```diff
@@ -31,6 +31,31 @@ export class TranslateStore {
...
+        }
+        this.loaders.set(loaderIndex, inject(TranslateLoader));
+        return loaderIndex;
+    }
+
+    removeLoader(loaderIndex: number) {
+        if (this.loaders.has(loaderIndex)) this.loaders.delete(loaderIndex);
+    }
+
+    getLoaders() {
```
CodeAndWeb: getLoaders() should deliver an array of **unique** loaders. 
we should hide the information that we use the indices. it's a implementation detail that does not need to be shared with the service.

using this:

~~~
provideTranslateService({
  loader: provideTranslateLoader(CustomLoader),
  compiler: provideTranslateCompiler(CustomCompiler)
})
~~~

~~~
  provideChildTranslateService({
    // no loader
  })
~~~

could add the same loader twice. this would lead to concurrent http requests for the same url.
this is why the list has to be unique.

projects/ngx-translate/src/lib/translate.service.ts
```diff
@@ -341,9 +349,22 @@ export class TranslateService implements ITranslateService {
     ): Observable<InterpolatableTranslationObject> {
         this.pending = true;
 
-        const loadingTranslations = this.currentLoader
-            .getTranslation(lang)
-            .pipe(shareReplay(1), take(1));
+        const loaders = Array.from(this.store.loaders.values());
```
CodeAndWeb: this should just be `this.store.getLoaders()`
I don't want the service to know about indices and implementation details in the store.

projects/ngx-translate/src/lib/translate.store.ts
```diff
@@ -28,6 +29,33 @@ export class TranslateStore {
...
+    /**
+     * Adds a new loader to the store
+     * @returns the index of the newly added loader
+     */
+    addLoader() {
+        let loaderIndex = 0;
+        while (this.loaders.has(loaderIndex)) {
+            loaderIndex++;
+        }
+        this.loaders.set(loaderIndex, inject(TranslateLoader));
```
CodeAndWeb: keep the `inject()` in the service. 
the service glues together the components. the store should not know where to get them from.

projects/ngx-translate/src/lib/translate.service.ts
```diff
@@ -185,6 +185,8 @@ export class TranslateService implements ITranslateService {
     private missingTranslationHandler = inject(MissingTranslationHandler);
```
CodeAndWeb, line 185: currentLoader is not used anymore. remove it.
it would only confuse users.

CodeAndWeb (changes requested):

CodeAndWeb:
... and finally, we need some tests for this functionality.

projects/ngx-translate/src/lib/translate.store.ts
rbalet: @CodeAndWeb but also `getLoaders()` is not useful, should I remove the both ? 

I've added getLoader, as I though it could be useful in the service for a refresh logic

projects/ngx-translate/src/lib/translate.store.ts
rbalet: @CodeAndWeb I do not understand the following can you explain ? 

> we should hide the information that we use the indices

_If you're talking about `_loaderIndex`, I though the service should be aware of it, as he's removing "himself" from the store on destroy_

------
2. Same loader added twice.
You're right, but I may be missing expertise here, do you now how to handle this possibility ? 

I though about generating a random `id` per loader so I could check this loader, but it seems to be wrong.   
But yet, we cannot get the instance ID in angular...

Doing so will force every use to add this value to theirs loaders. 

But I do now understand why you wanted a counter, as it could be casted multiple time, you'd like to add, remove and delete it when it gets to 0. 

Am I right ?

rbalet:
> const loaders = Array.from(this.store.loaders.value

Yes, I'll do the testing once we're done with the changes ;)

projects/ngx-translate/src/lib/translate.service.ts
rbalet: Yes, I'll do once I'm doing the testing

rbalet:
@CodeAndWeb Thx for the help 😋

rbalet:
@CodeAndWeb Can you notify me once an RC is released, that I can test it out on some projects? 

Cheers

PR #1600, opened 2025-10-09
fix: make private properties and methods protected in order to allow usage in extending classes

reey:
We are extending some of these classes for our usecases. Some of the properties and methods were previously public and are not private. Making the protected would allow us to access them again in our extending classes..


CodeAndWeb:
I'll implement it on the new update.

Just be warned: Things might still change a lot internally... we try to keep the public interface stable as good as possible - but do not take any precautions for the protected/private stuff.

CodeAndWeb:
=> 34199416b15093da9d485ee888a1a1ddc07cb4ce

Issue #1606, opened 2025-12-17
Re-introduce a way to get translations from other languages than current


OhSoooLucky:
## Current behavior
Currently, most notably with the recent changes. There's no way to get translations for another language without having to change the current language.

## Expected behavior
Even though it might not always be used, there's no reason to block the functionality of translating other languages other than the currently selected language.

## What is the motivation / use case for changing the behavior?
We at Group Claes adopted the deprecated @groupclaes/ngx-translate-router project from Gilsdav. We're updating the plugin to handle SSR and especially SSG routes. Currently there's no exposed way to load translations on the server-side to generate and translate the routes for languages other than the current language.

## How do you think that we should implement this?
Provide overloads to read based on the key and language for translateservice.instant or translateservice.get

pweix:
I also have this problem, i now have the problem, that my custom pipe to get a single translated value of a specified language by key doesn't work anymore. The code used is:

```
return this.ts.getTranslation(iso).pipe(map(translations => {
        return this.ts.getParsedResult(translations, key, interpolateParams);
}));
```
Field `this.ts` is my custom service extending `TranslateService`. Unfortunately `getTranslation` doesn't exist anymore and `getParsedResult` only takes two parameters now (passing translations is not supported anymore apparently). Searched all documentation for a replacement, but couldn't find anything. 

Stuck on v16 for now -.-

sebvie:
We have the same problem. Our use case is to have a text pipe in 2 languages at the same time:

welcome / Willkommen


our workaround is:

```
      const prev = this.translate.getCurrentLang();
      return this.translate.use(lang).pipe(
        switchMap(() => this.translate.get(key, params)),
        finalize(() => {
          this.translate.use(prev);
        })

```

CodeAndWeb:
This is implemented on `main` in commit ngx-translate/core@5bc8773 — `get()`, `stream()`, and `instant()` now all accept an optional `lang` parameter that bypasses the current language and fallback chain, which should cover the SSR/SSG use case. It'll ship in the upcoming 18.0.0 release.

PR #1607, opened 2025-12-17
feat(service/store): Allow translations for any supported language using TranslateService

OhSoooLucky:
This PR allows users to fetch a translation for any language

Current To-Do's/remarks:
- [ ] The language must first be pre-loaded using loadTranslations()
- [ ] TranslateService: Check if the language is a supported language
- [x] Write unit tests

Solution for issue #1606


CodeAndWeb:
I've added support for this to v18

Issue #1609, opened 2026-01-12
interpolating an array's length is not working in v17 anymore


soletan:
## Current behavior

When using interpolation patterns like `{{ matches.length }}` the resulting string does not contain the length of array `matches`, but that pattern's source as string.

## Expected behavior

It should be possible to inject an array's size into translations, e.g. given the pattern above used in a translation for a key 'FOO.BAR' should result in that array's length given as number.

## How do you think that we should fix this?

Method getValue() in util.ts has changed in v17 and it is limiting support for descending into arrays accepting numeric keys, only. Hence, using `.length` is not working like before anymore.

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/github-acm14mj2?file=src%2Fapp%2Fapp.component.ts

<img width="1251" height="423" alt="Image" src="https://github.com/user-attachments/assets/d4a3867a-9bc5-4634-bf7d-20e01429d0d4" />

## Environment

<pre><code>
ngx-translate version: 17.0.0
Angular version: 21.0.8

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 22.20.0  
- Platform:  Windows 

Others:

</code></pre>

soletan:
Any chance this is going to be available in a release of yours anytime soon or do we have to set up a fork and publish the fix (for a regression issue in your code) ourselves?

CodeAndWeb:
Thanks for the detailed bug report and the stackblitz repro, @soletan! 

Fixed on `develop` in 0bff987 (will ship in v18):

- `.length` now resolves as a terminal lookup on array interpolation params.
- Numeric-index lookups are now gated behind a `/^\d+$/` regex, so malformed keys like `"1x"` are no longer silently resolved via `parseInt` leading-digit parsing.
- Other array properties (`constructor`, `toString`, `push`, etc.) remain intentionally inaccessible — the contract is tighter than pre-regression.

No fork needed. v18 is imminent.

PR #1610, opened 2026-01-12
fix: feature array length on interpolation

soletan:
On interpolating arrays, using `length` results in the array's length as this has not been working anymore before since v17.


soletan:
related to #1609

CodeAndWeb:
can you please also add a test for this?

soletan:
unit tests have been added

Nosfistis:
Hello @CodeAndWeb is it possible to have this be merged and released soon? It is an important regression for us.

CodeAndWeb:
it's part of v18

PR #1623, opened 2026-06-11
fix(core): detect nameless class expressions in isClass helper

### LINKED ISSUE #1622: Application breaks if provided class is nameless in build output
## Current behavior

When `optimize.scripts` is `true` in the Angular CLI config, my application does not load due to `TypeError: Class constructor Oe cannot be invoked without 'new'`

Image

## Expected behavior

The application loads.

In https://github.com/ngx-translate/core/blob/d42c87169c4285acbc67b0f98a93874b6c9b36bd/projects/ngx-translate/src/lib/translate.providers.ts#L43 extend the Regex from `class\s` to `class[{\s]`

This way it also handles minified nameless classes, which can look like:

```ts
Oe=class{de=se$1(se$1({},Jt),Xt);en=se$1(se$1({},Qt),qt);getTranslation(l){return M(l==="de"?this.de:this.en)}}
```

## Environment

<pre><code>
ngx-translate version: 18.0.0
Angular version: 22.0.0

Browser:
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [x] Edge version 149
 
For Tooling issues:
- Node version: -  
- Platform: - 

Others:

</code></pre>

jonaskuske:
Closes #1622

### DIFF
--- a/projects/ngx-translate/src/lib/translate.providers.ts
+++ b/projects/ngx-translate/src/lib/translate.providers.ts
@@ -40,7 +40,9 @@ export interface RootTranslateServiceConfig extends ChildTranslateServiceConfig 
 }
 
 function isClass<T>(fn: Type<T> | (() => T)): fn is Type<T> {
-    return /^class\s/.test(Function.prototype.toString.call(fn));
+    // for declarations, "class" is followed by \s: class Foo {...
+    // for expressions, "class" can be unnamed and followed by {: foo = class{...
+    return /^class[{\s]/.test(Function.prototype.toString.call(fn));
 }
 
 function toProvider<T>(


CodeAndWeb:
Thanks for this, @jonaskuske — your diagnosis was spot on and your `/^class[{\s]/` fix is exactly right.

I'm closing this in favour of #1624 (now merged), which makes the identical one-line change plus adds a regression test that reproduces the minified `class{…}` output via `eval` (a class literal written in source gets re-indented to `class … {` by Prettier and wouldn't reproduce the bug). Same fix, just with a guard so it can't regress.

Really appreciate you taking the time to track this down and send a PR. 🙏

PR #1663, opened 2026-09-22
fix(core): don't let a nested key shadow a flat sibling key in getValue

### LINKED ISSUE #1561: TranslateService.set breaks the localization keys with . inside
## Current behavior

In our application we provide our localization with the following format 
```json
{
   "my-component-selector.key1": "value1",
   "my-component-selector.key2": "value2"
}
```

If I call `TranslateService.set('my-component-selector.key1', 'new value')`.
It updates correctly `my-component-selector.key1` but I lost the localization of `my-component-selector.key2`.
From what I understood it is because now localization keys are handled with hierarchy.
So after the update we should have something like:
```json
{
  "my-component-selector": {
     "key1": "new value"
   },
   "my-component-selector.key2": "value2"
}
```
and now the translate service for `my-component-selector.key2` returns `undefined`.

## Expected behavior

I would expect `my-component-selector.key2` to still be translated correctly as it works before the call to [`TranslateService.set`](https://github.com/ngx-translate/core/blob/develop/projects/ngx-translate/src/lib/translate.service.ts#L512)

I think 2 options are possible here:
- The resolution of keys should first check the hierarchy as for now, and if `undefined` then check for the key as is.
- `TranslateService.set` could take an extra parameter to keep the format, in the example `my-component-selector.key1`  instead of `{my-component-selector: {key1: 'new value'}}`

- Define differents localization keys with the same prefix and `.` as separator.
- Call `TranslateService.set`

## Environment

<pre><code>
ngx-translate version: 16.0.0
Angular version: 19.0.0

Browser:
- [x] Chrome (desktop) version XX
- [x] Chrome (Android) version XX
- [x] Chrome (iOS) version XX
- [x] Firefox version XX
- [x] Safari (desktop) version XX
- [x] Safari (iOS) version XX
- [x] IE version XX
- [x] Edge version XX
 
For Tooling issues:
- Node version: 20 || 22  
- Platform:   Mac, Linux, Windows

Others:
N/A
</code></pre>

OsamaAnsar:
## Summary

`TranslateService.set('a.b.key1', value)` on a flat, dot-containing translation key silently breaks lookup of every sibling key sharing that prefix (e.g. `a.b.key2`).

- `TranslateService.set()` ([`translate.service.ts:959`](https://github.com/ngx-translate/core/blob/develop/projects/ngx-translate/src/lib/translate.service.ts#L959)) calls `insertValue()`.
- `insertValue()` ([`util.ts:194`](https://github.com/ngx-translate/core/blob/develop/projects/ngx-translate/src/lib/util.ts#L194)) always builds a *nested* object for the dotted key (via `createNestedObject`) and deep-merges it into the translations object. So after `set('a.b.key1', ...)` on translations that started as flat keys (`{ "a.b.key1": "v1", "a.b.key2": "v2" }`), the object ends up with **both** the original flat keys and a brand-new `{ a: { b: { key1: ... } } }` structure sharing the same prefix.
- `getValue()` ([`util.ts:137`](https://github.com/ngx-translate/core/blob/develop/projects/ngx-translate/src/lib/util.ts#L137)) resolves a dotted key by progressively matching segments, and greedily descends into a dict as soon as it finds one for the current segment — regardless of whether the rest of the path actually resolves inside it. So looking up the *untouched* sibling `a.b.key2` walks into the newly created `a.b` nested object, doesn't find `key2` there, and returns `undefined`, even though `a.b.key2` is still sitting on the object, unmodified, as a flat key.

This matches #1561, reported against v16.0.0: calling `set()` on one dotted key corrupts lookup of every sibling flat key sharing its prefix.

Fixes #1561

## Fix

`getValue()` now falls back to a literal lookup of the *full* dotted key on the original object when the progressive/nested lookup comes up empty, instead of giving up and returning `undefined`. This fallback only triggers when the normal lookup found nothing, so:
- Keys that already resolve correctly today (nested-only objects, or keys where the nested path does resolve) are completely unaffected — nested resolution still takes priority, matching current behavior when both a nested and flat form exist and the nested one resolves.
- The fix is scoped to `getValue()` only; `insertValue()`/`createNestedObject()`/`set()` are untouched.

## Test plan

Added a regression test at both levels:
- `projects/ngx-translate/src/tests/utils.spec.ts` (`getValue()` describe block): seeds `{ "a.b.key1": "v1", "a.b.key2": "v2" }`, runs it through `insertValue()` the same way `set()` does, and asserts `getValue()` still resolves `a.b.key2` after `a.b.key1` was "set".
- `projects/ngx-translate/src/tests/translate.service.spec.ts` (`set()` describe block): seeds the same two flat sibling keys via `setTranslation`, calls `translate.set('a.b.key1', 'newV1', 'en')`, and asserts `translate.instant('a.b.key2')` still returns `'v2'` (not `undefined`) while `a.b.key1` correctly reflects the new value.

Verified:
- [x] Both new tests fail with the exact reported symptom (`a.b.key2` → `undefined`) when the `getValue()` fix is reverted and only the tests are kept — confirmed by temporarily stashing the source change, re-running (`2 FAILED, 431 SUCCESS`), then restoring the fix and re-running (`433 SUCCESS`).
- [x] Full `ngx-translate` test suite passes: `TOTAL: 433 SUCCESS` (`ng test ngx-translate --watch=false --browsers=ChromeHeadless`), including all existing `getValue()`/`insertValue()`/nested-object tests (e.g. `should set translations with nested key (object)`, and the existing flat-vs-nested ambiguity cases in `utils.spec.ts`).
- [x] `ng lint` passes for all three projects (`ngx-translate`, `http-loader`, `test-app`).
- [x] `tsc --noEmit -p tsconfig.spec.json` passes with no errors.
- [x] `prettier --check` passes on the changed files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rbalet:
@OsamaAnsar Thanks for the PR. 
Could you please open an issue and explain there your problem ? 

That is really hard to read so much AI generated text. 
Cheers

rbalet (changes requested):
Please, reduce the total amount of comment and use /* */ for multiline comment instead

